PR #5, opened 2017-09-18
Added bin folder to path

nop90:


crt0/Makefile
```diff
@@ -1,8 +1,10 @@
+export PATH	:=	$(DEVKITA64)/bin:$(PATH)
+
 switch_crt0.o: *.s
 	aarch64-none-elf-gcc -march=armv8-a -x assembler-with-cpp -c $^ -o $@
 
 install: switch_crt0.o
-	cp $^ $(DEVKITA64)/aarch64-none-elf/lib/$^
+	cp $^ $(DEVKITA64)/bin/aarch64-none-elf/lib/$^
```
d3m3vilurr: it's probably wrong.

nop90: you're right, 

that was from the wrong file. Sorry.

PR #18, opened 2017-11-08
devkitA64 alpha 5 updates (locks, TLS, etc) {TEST BEFORE MERGE}

fincs:
- [x] Using `-z nodynamic-undefined-weak` in linker options to remove dynamic lookup of undefined weak symbols (reduces unneeded bloat in generated binaries)
- [x] Add `-mtp=soft` to ARCH setting in makefile (*this change needs to be done in all projects including the switch examples*, I'll PR that later)
- [x] Integrate newlib locks with libnx (R)Mutex code
- [x] Add newlib reent support to libnx threads
- [x] Add TLS segment support to libnx threads

All of this is now done, however it's completely untested by me, since I can't run switch code (yet?)
Please test before merging.

### DIFF
--- a/nx/source/kernel/thread.c
+++ b/nx/source/kernel/thread.c
@@ -1,9 +1,32 @@
 // Copyright 2017 plutoo
 #include <switch.h>
 #include <malloc.h>
+#include "../internal.h"
 
-static void _EntryWrap(Thread* t) {
-    t->entry(t->arg);
+extern const u8 __tdata_lma[];
+extern const u8 __tdata_lma_end[];
+extern u8 __tls_start[];
+extern u8 __tls_end[];
+
+typedef struct {
+    Thread*        t;
+    ThreadFunc     entry;
+    void*          arg;
+    struct _reent* reent;
+    void*          tls;
+    void*          padding;
+} ThreadEntryArgs;
+
+static void _EntryWrap(ThreadEntryArgs* args) {
+    // Initialize thread vars
+    ThreadVars* tv = getThreadVars();
+    tv->magic      = THREADVARS_MAGIC;
+    tv->thread_ptr = args->t;
+    tv->reent      = args->reent;
+    tv->tls_tp     = (u8*)args->tls-2*sizeof(void*); // subtract size of Thread Control Block (TCB)
+
+    // Launch thread entrypoint
+    args->entry(args->arg);
     svcExitThread();
 }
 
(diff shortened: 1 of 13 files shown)


nx/source/system/readtp.s
```diff
@@ -0,0 +1,10 @@
+	.section .text.__aarch64_read_tp, "ax", %progbits
+	.global __aarch64_read_tp
+	.type __aarch64_read_tp, %function
+	.align 2
+	.cfi_startproc
+__aarch64_read_tp:
```
plutooo: Why is this in assembly

nx/source/kernel/thread.c
```diff
@@ -1,18 +1,45 @@
...
+    // Launch thread entrypoint
+    args->entry(args->arg);
     svcExitThread();
 }
 
 Result threadCreate(
     Thread* t, ThreadFunc entry, void* arg, size_t stack_sz, int prio,
     int cpuid)
 {
+    stack_sz = (stack_sz+0xF) &~ 0xF;
```
plutooo: Needs 4K align

nx/source/kernel/thread.c
```diff
@@ -24,21 +51,41 @@ Result threadCreate(
 
         if (R_SUCCEEDED(rc))
         {
-            u64 stack_top = ((u64)stack_mirror) + t->stack_sz;
+            u64 stack_top = ((u64)stack_mirror) + t->stack_sz - sizeof(ThreadEntryArgs);
```
plutooo: Must be 16-aligned

nx/source/system/readtp.s
fincs: It needs to be, __aarch64_read_tp is not following standard calling conventions (neither was __aeabi_read_tp in 32-bit ARM).

nx/source/kernel/thread.c
fincs: It already is, ThreadEntryArgs has an extra member to keep 16-byte alignment.

nx/source/kernel/thread.c
fincs: Will be fixed.

nx/source/system/readtp.s
plutooo: OK

nx/source/kernel/thread.c
plutooo: Yeah, but this can explode later when an innocent looking change breaks everything xD

fincs: Should I add a comment to ThreadEntryArgs in case it needs to be changed in the future?

plutooo: Yeah, or you could add code to align

plutooo:
LGTM, not tested

nx/source/services/fatal.c
```diff
@@ -36,4 +36,5 @@ void fatalSimple(Result err) {
     }
 
     ((void(*)())0xBADC0DE)();
+	__builtin_unreachable();
```
d3m3vilurr: mixed hardtab & softtab.

fincs: Whoops

PR #19, opened 2017-11-13
Elf->NRO support

shinyquagsire23:
Loads correctly with Pegaswitch, about as bare-minimum as elf2nso (no relocation/linking to NSOs, just loads in the sections and sets header info correctly). Probably doesn't work outright with crt0 (nor should it, tbh) but it's useful for building loadable code with devkitA64+libnx.


plutooo:
LGTM, been talking with people on how to resolve the crt0 issue

PR #20, opened 2017-11-14
elf2nro tweaks

shinyquagsire23:
Did some more testing (and forgot to commit some code...), the RX+R+RW sections are guaranteed to be located in sequence, so it's best to use the vaddr as the file writing location. ~~With the current crt0, however, this causes the NRO0 header to overwrite some of .text, can be avoided with some padding though.~~ crt0 has been adjusted to account for the NRO header.

### DIFF
--- a/crt0/switch_crt0.s
+++ b/crt0/switch_crt0.s
@@ -3,8 +3,10 @@
 
 _start:
     bl startup
+    .word 0
     .ascii "HOMEBREW"
 
+.org _start+0x80
 startup:
     // get aslr base
     sub  x28, x30, #4
--- a/tools/elf2nro.c
+++ b/tools/elf2nro.c
@@ -105,10 +105,9 @@ int main(int argc, char* argv[]) {
 
     Elf64_Phdr* phdrs = (Elf64_Phdr*) &elf[hdr->e_phoff];
     size_t i, j = 0;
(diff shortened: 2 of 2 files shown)


fincs:
Any idea why that happens? Either elf2nro or crt0/ldscript should be fixed.

plutooo:
I'm thinking since first u32 of NRO0 is unused, we can put a jump instruction there in case it runs as NSO0.

So we can have a hybrid crt0

shinyquagsire23:
yeah ldr:ro seems to require that .text start at +0 in the file, so the NRO header has to be in .text. I'll see if ldr:ro is happy with there being a branch instruction in the NRO header and if it is then I think crt0 can just be adjusted to pad out to 0x80 for the actual start.

crt0/switch_crt0.s
```diff
@@ -3,8 +3,10 @@
 
 _start:
     bl startup
+    .word 0
     .ascii "HOMEBREW"
 
+.org 0x80
```
fincs: Strictly speaking this should either be `.org _start+0x80`, or using a `.space` directive.

shinyquagsire23:
It's totally fine with there being a branch instruction, along with the HOMEBREW tag in the other bit of padding.

plutooo:
Sweet

PR #22, opened 2017-11-20
HID Shared Memory - Structures + key held/down/up implementation

shinyquagsire23:
Example of usage:
```
// Retrieve only raw left output from Player 1
hidSetControllerLayout(CONTROLLER_PLAYER_1, LAYOUT_LEFT);
while(1)
{
    hidScaninput();

    if (hidKeysHeld(CONTROLLER_PLAYER_1) & KEY_DOWN)
        printf("P1 left controller is held down\n");

    if (hidKeysDown(CONTROLLER_HANDHELD) & KEY_A)
        printf("Handheld pressed A!\n");

    if ((hidMouseButtonsHeld() & MOUSE_LEFT) && hidKeyboardHeld(KBD_W))
        printf("Pyro main detected!\n");

    if (hidKeyboardModifierUp(KBD_MOD_CAPSLOCK))
        printf("caps lock finally released\n");

    if (hidKeysDown(CONTROLLER_HANDHELD) & KEY_TOUCH) {
        touchPosition pos;
        printf("Touches at:\n");
        for (int i = 0; i < hidTouchCount(); i++) {
            hidTouchRead(&pos, i);
            printf("Touch index #%u: %u, %u\n", i, pos.px, pos.py);
        }
    }
}
```

Keyboard keys are handled with the key passed in due to the keyboard keys being held in a 256-bit array, a bit per key based on the HID scancode. Scancodes for keyboards cannot be OR'd together (two separate hidKeyboard* calls OR'd works fine however). Keyboard modifiers can technically be handled the same as controller input, but I had them passed in as arguments for consistency.

Controller input can be retrieved from several different types of layouts, the default layout is `LAYOUT_DEFAULT` which adjusts for controller orientation for Joy-Con halves. This can be overridden to another layout if desired, per-controller.

HID shared memory stores timestamps in number of samples, controller updates are checked for by subtracting the current timestamp from the last timestamp to check if samples have been added. Samples are stored in circular buffers with `latestEntry` indexing the current sample entry. The current behavior follows ctrulib where current samples are cleared and are only set to not-0 if there are updated samples available at the time of calling `hidScanInput`

Touchscreen structures are implemented, I'm still debating exactly how to handle/store held/down/up there since the number of fingers used is variable. More status probing stuff is still incoming as well, but again all structures are implemented for the most part.

- [x] Keys held/down/up for mouse, keyboard, controllers
- [ ] Controller status probing (which controllers are active, what type of controllers are they, how many, what colors are they, what layouts are they using)
- [x] Touchscreen value retrieval, and KEY_TOUCH
- [ ] Activate/deactivate input devices, only probe what is active

### DIFF
--- a/nx/source/services/hid.c
+++ b/nx/source/services/hid.c
@@ -110,6 +110,10 @@ void hidSetControllerLayout(HIDControllerID id, HIDControllerLayoutType layoutTy
     g_controllerLayout[id] = layoutType;
 }
 
+HIDControllerLayoutType hidGetControllerLayout(HIDControllerID id) {
+    return g_controllerLayout[id];
+}
+
 void hidScanInput(void) {
     if (g_hidServiceSession == INVALID_HANDLE) return;
     HIDSharedMemory *sharedMem = (HIDSharedMemory*)hidGetSharedmemAddr();
--- a/nx/source/services/hid.c
+++ b/nx/source/services/hid.c
@@ -137,6 +137,9 @@ void hidScanInput(void) {
     if ((s64)(newTouchEntry->header.timestamp - g_touchTimestamp) > 0) {
         memcpy(&g_touchEntry, newTouchEntry, sizeof(HIDTouchScreenEntry));
         g_touchTimestamp = newTouchEntry->header.timestamp;
+
+        if (hidTouchCount())
+            g_controllerHeld[CONTROLLER_HANDHELD] |= KEY_TOUCH;
     }
 
     u64 latestMouseEntry = sharedMem->mouse.header.latestEntry;
@@ -176,7 +179,7 @@ void hidScanInput(void) {
             memcpy(&g_controllerEntries[i], newInputEntry, sizeof(HIDControllerInputEntry));
             g_controllerTimestamps[i] = newInputEntry->timestamp;
 
-            g_controllerHeld[i] = g_controllerEntries[i].buttons;
+            g_controllerHeld[i] |= g_controllerEntries[i].buttons;
         }
 
         g_controllerDown[i] = (~g_controllerOld[i]) & g_controllerHeld[i];
@@ -236,6 +239,25 @@ bool hidKeyboardDown(HIDKeyboardScancode key) {
 
 bool hidKeyboardUp(HIDKeyboardScancode key) {
     return g_keyboardUp[key / 32] & (1 << (key % 32));
+}
+
+u32 hidTouchCount(void) {
+    return g_touchEntry.header.numTouches;
+}
+
+void hidTouchRead(touchPosition *pos, u32 point_id) {
+    if (pos) {
+        if (point_id >= g_touchEntry.header.numTouches) {
+            memset(pos, 0, sizeof(touchPosition));
+            return;
+        }
+
+        pos->px = g_touchEntry.touches[point_id].x;
+        pos->py = g_touchEntry.touches[point_id].y;
+        pos->dx = g_touchEntry.touches[point_id].diameterX;
+        pos->dy = g_touchEntry.touches[point_id].diameterY;
+        pos->angle = g_touchEntry.touches[point_id].angle;
+    }
 }
 
 static Result _hidCreateAppletResource(Handle sessionhandle, Handle* handle_out, u64 AppletResourceUserId) {
--- a/nx/source/services/hid.c
+++ b/nx/source/services/hid.c
@@ -5,6 +5,19 @@ static Handle g_hidServiceSession = INVALID_HANDLE;
 static Handle g_hidIAppletResource = INVALID_HANDLE;
 static SharedMemory g_hidSharedmem;
 
+static HIDTouchScreenEntry touchEntry;
+static HIDMouseEntry mouseEntry;
+static HIDKeyboardEntry keyboardEntry;
+static HIDControllerInputEntry controllerEntries[10];
+
+static u64 mouseOld, mouseHeld, mouseDown, mouseUp;
+static u64 keyboardModOld, keyboardModHeld, keyboardModDown, keyboardModUp;
+static u32 keyboardOld[8], keyboardHeld[8], keyboardDown[8], keyboardUp[8];
+static u64 controllerOld[10], controllerHeld[10], controllerDown[10], controllerUp[10];
+
+static HIDControllerLayoutType controllerLayout[10];
+static u64 touchTimestamp, mouseTimestamp, keyboardTimestamp, controllerTimestamps[10];
+
 static Result _hidCreateAppletResource(Handle sessionhandle, Handle* handle_out, u64 AppletResourceUserId);
 static Result _hidGetSharedMemoryHandle(Handle sessionhandle, Handle* handle_out);
 
@@ -41,6 +54,8 @@ Result hidInitialize(void) {
         g_hidIAppletResource = INVALID_HANDLE;
     }
 
+    hidReset();
+
     return rc;
 }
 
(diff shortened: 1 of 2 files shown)


nx/source/services/hid.c
```diff
@@ -61,6 +76,29 @@ void hidExit(void)
     shmemClose(&g_hidSharedmem);
 }
 
+void hidReset(void)
+{
+    // Reset internal state
+    memset(&touchEntry, 0, sizeof(HIDTouchScreenEntry));
+    memset(&mouseEntry, 0, sizeof(HIDMouseEntry));
+    memset(&keyboardEntry, 0, sizeof(HIDKeyboardEntry));
+    memset(&controllerEntries, 0, sizeof(HIDControllerInputEntry));
```
yellows8: This is an array not a single entry: remove '&' and use sizeof(controllerEntries).

nx/source/services/hid.c
```diff
@@ -69,6 +107,149 @@ void* hidGetSharedmemAddr(void) {
...
+    memcpy(controllerOld, controllerHeld, sizeof(controllerOld));
+
+    mouseHeld = 0;
+    keyboardModHeld = 0;
+    memset(keyboardHeld, 0, sizeof(keyboardHeld));
+    memset(controllerHeld, 0, sizeof(controllerHeld));
+    memset(&touchEntry, 0, sizeof(HIDTouchScreenEntry));
+    memset(&mouseEntry, 0, sizeof(HIDMouseEntry));
+    memset(&keyboardEntry, 0, sizeof(HIDKeyboardEntry));
+    for (int i = 0; i < 10; i++)
```
yellows8: Replace with a memset for the entire array as described above.

nx/source/services/hid.c
```diff
@@ -5,6 +5,19 @@ static Handle g_hidServiceSession = INVALID_HANDLE;
 static Handle g_hidIAppletResource = INVALID_HANDLE;
 static SharedMemory g_hidSharedmem;
 
+static HIDTouchScreenEntry touchEntry;
```
plutooo: Can you prefix with g_?

nx/source/services/hid.c
```diff
@@ -61,6 +76,28 @@ void hidExit(void)
     shmemClose(&g_hidSharedmem);
 }
 
+void hidReset(void) {
+    // Reset internal state
+    memset(&touchEntry, 0, sizeof(HIDTouchScreenEntry));
```
plutooo: You can turn all of this into an anonymous struct and then just have a single memset-call if you want to

fincs:
How about:
- `u32 hidTouchCount(void)` returning the number of touch points
- `void hidTouchRead(touchPosition* pos, u32 point_id)` for reading a touch point (0 is always known to exist)
- `KEY_TOUCH` pseudo-key that would be added to the key bitmask when `hidTouchCount() != 0`

plutooo:
How to know what controller configuration is active?

nx/source/services/hid.c
```diff
@@ -69,6 +106,135 @@ void* hidGetSharedmemAddr(void) {
...
+    memset(controllerHeld, 0, sizeof(controllerHeld));
+    memset(&touchEntry, 0, sizeof(HIDTouchScreenEntry));
+    memset(&mouseEntry, 0, sizeof(HIDMouseEntry));
+    memset(&keyboardEntry, 0, sizeof(HIDKeyboardEntry));
+    for (int i = 0; i < 10; i++)
+        memset(&controllerEntries[i], 0, sizeof(HIDControllerInputEntry));
+
+    u64 latestTouchEntry = sharedMem->touchscreen.header.latestEntry;
+    HIDTouchScreenEntry *newTouchEntry = &sharedMem->touchscreen.entries[latestTouchEntry];
+    if (newTouchEntry->header.timestamp - touchTimestamp) {
```
plutooo: What's going on with this if statement

nx/source/services/hid.c
```diff
@@ -69,6 +106,135 @@ void* hidGetSharedmemAddr(void) {
     return shmemGetAddr(&g_hidSharedmem);
 }
 
+void hidSetControllerLayout(HIDControllerID id, HIDControllerLayoutType layoutType) {
+    controllerLayout[id] = layoutType;
+}
+
+void hidScanInput(void) {
```
plutooo: I think we might wanna either split hidScanInput into:

hidScanKeyboard()
hidScanController()
hidScanTouch()
hidScanMouse()
etc...

or have it take a bitfield like this:

hidScanInput(HidScan_Controller | HidScan_Touch)

plutooo: For performance reasons

yellows8:
(De)activate service commands should be added too. (http://switchbrew.org/index.php?title=HID_services#hid)

nx/source/services/hid.c
yellows8: Or only process the currently activated devices?

fincs: I think controller state (and touch?) should be unconditionally scanned. The other devices (keyboard, mouse) should only be scanned if they are active/enabled. I don't think `hidScanInput` should be split into multiple functions, or a parameter added to it.

plutooo: Why have a special rule? It's better to have the API symmetric for all types of input

nx/source/services/hid.c
fincs: I think it should be rewritten to `if ((s64)(newTouchEntry->header.timestamp - touchTimestamp) >= 0)`. It is meant to be a basic greater-than check, accounting for the fact that the timestamp overflowing the 64-bit variable is technically possible and valid.

nx/source/services/hid.c
fincs: But not all types of input will be as commonly used. I think button input should always be available.

nx/source/services/hid.c
plutooo: Right now the comparison is equivalent to (newTouchEntry->header.timestamp != touchTimestamp) afaict

fincs: It is. But of course, incoming touch entries with a previous timestamp would also pass the check... Not sure if we want to assume that will never happen...

nx/source/services/hid.c
plutooo: Doesn't justify a special rule.

A function should not assume what the user wants to do. It should do what it says, and only that.

fincs: Actually, is it possible to disable controller input in the first place? If so, then there wouldn't really be a way to unconditionally disable button input.

I think libnx could keep track of which sources of input are enabled at the moment, and scan those in `hidScanInput`, therefore obviating the need for a special parameter.

nx/source/services/hid.c
shinyquagsire23: Rollover is unlikely, but I think I'll go ahead and move to a greater-than zero compare for clarity.

nx/source/services/hid.c
shinyquagsire23: I'm thinking what I'll do is add the activate/deactivate commands @yellows8 mentioned, keep keyboard+mouse off by default at init and then have libnx keep track of and only update button states for devices which are enabled.

I think I can also optimize controllers a bit by checking if the slot has no controller attached, HIDControllerType should have no bits set with controllers detached. I'm not sure if I have anything similar to detect mouse+keyboard though, I'll have to investigate a bit I suppose.

shinyquagsire23:
Went ahead and added a list of things yet to be added, I think at the very least I'd like feature parity with ctrulib w/ touch stuff (maybe minus gyro stuff, need to look into that). Might also look into organizing things into structures some more.

plutooo:
Let's just pull this in

PR #24, opened 2018-01-09
vi:m ManagedLayer support

shinyquagsire23:
vi:m is able to create its own managed layers without AppletAE/AppletAO. Also fixed a typo in gfx.c.


plutooo:
Can you remove the logic from viOpenLayer so that we can merge this ?

PR #34, opened 2018-01-22
Make the bsd service work properly

TuxSH:
This fixes the initalization of the BSD service(s). Without this commit, the bsd services wouldn't send or receive more than 1 byte/s using TCP, as it was issuing ZeroWindow (unbuffered) packets, symptom of a lack of memory.

The following helper functions have been added:

* ~~`bsdGetTransferMemSizeForConfig`, returning the minimal size needed for the transfer memory to be passed to `bsdInitalize` (using any size less than that will cause the abovementionned issue)~~
* `bsdGetDefaultConfig`, returning a reasonable default configuration for the developer to use
* `bsdInitalizeDefault`

The `bsd` service code now manages the transfer memory itself.
`bsdExit` has been added.

Some const-correctness issues have also been made.

### DIFF
--- a/nx/source/services/bsd.c
+++ b/nx/source/services/bsd.c
@@ -7,13 +7,17 @@
 #include <switch/kernel/shmem.h>
 #include <switch/kernel/rwlock.h>
 
+#define EPIPE 32
+
 static Service g_bsdSrv;
 static Service g_bsdMonitor;
 static u64 g_bsdClientPid = -1;
 static int g_Errno = 0;
 
+static TransferMemory g_bsdTmem;
+
 static const BsdConfig g_defaultBsdConfig = {
-    .version = 2,
+    .version = 1,
 
     .tcp_tx_buf_size        = 0x8000,
     .tcp_rx_buf_size        = 0x10000,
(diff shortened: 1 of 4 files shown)


nx/source/services/bsd.c
```diff
@@ -12,9 +12,23 @@ static Service g_bsdMonitor;
 static u64 g_bsdClientPid = -1;
 static int g_Errno = 0;
 
+static const BsdConfig g_defaultBsdConfig = {
+    .version = 2,
```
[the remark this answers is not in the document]
TuxSH: Seems unchecked by the BSD services, but when I tested 1 works as well, yeah.
Should use 1 indeed.

nx/source/services/bsd.c
```diff
@@ -92,7 +104,11 @@ static Result _bsdStartMonitor(Service* srv, u64 pid) {
     return rc;
 }
 
-Result bsdInitialize(TransferMemory* tmem) {
+const BsdConfig *bsdGetDefaultConfig(void) {
+    return &g_defaultBsdConfig;
+}
+
+Result bsdInitialize(TransferMemory* tmem, const BsdConfig *config) {
```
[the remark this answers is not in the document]
TuxSH: Was already like that -- will make the appropriate changes if you want me to. Seems cleaner indeed.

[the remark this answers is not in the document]
plutooo: Yeah it was a mistake with the original code

nx/source/services/bsd.c
```diff
@@ -92,30 +121,54 @@ static Result _bsdStartMonitor(Service* srv, u64 pid) {
...
+    rc = _bsdRegisterClient(&g_bsdSrv, &g_bsdTmem, config, &g_bsdClientPid);
+    if(R_FAILED(rc)) goto error;
 
+    rc = _bsdStartMonitor(&g_bsdMonitor, g_bsdClientPid);
+    if(R_FAILED(rc)) goto error;
+
+    return rc;
+
+error:
+    if(serviceIsActive(&g_bsdMonitor))
```
plutooo: serviceClose() was designed so that you don't need to check if it's active

nx/source/services/bsd.c
```diff
@@ -92,30 +121,54 @@ static Result _bsdStartMonitor(Service* srv, u64 pid) {
...
+        serviceClose(&g_bsdMonitor);
+    if(serviceIsActive(&g_bsdSrv))
+        serviceClose(&g_bsdSrv);
+    if(&g_bsdTmem.src_addr != NULL)
+        tmemClose(&g_bsdTmem);
```
plutooo: ditto with tmemClose()

nx/source/services/bsd.c
TuxSH: Agree

nx/source/services/bsd.c
TuxSH: I don't agree, INVALID_HANDLE is -1 and not 0 (and g_bsdTmem->handle == 0 if it has never been initialized)

plutooo: Yeah -- it's a bug in tmem

PR #39, opened 2018-01-30
Restore initial sp on application exit

TuxSH:
Note: untested.

### DIFF
--- a/nx/source/runtime/switch_crt0.s
+++ b/nx/source/runtime/switch_crt0.s
@@ -33,6 +33,11 @@ bss_loop:
     subs x1, x1, #8
     bne  bss_loop
 
+    // store stack pointer
+    mov  x1, sp
+    adrp x0, __stack_top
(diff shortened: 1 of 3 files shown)


nx/source/runtime/switch_crt0.s
```diff
@@ -33,6 +33,11 @@ bss_loop:
     subs x1, x1, #8
     bne  bss_loop
 
+    // store stack pointer
+    mov  x1, sp
+    adrp x0, __stack_top
+    str  x1, [x0, #:lo12:__stack_top]
```
plutooo: Does the str instruction support up to 12-bit offsets?

fincs: This is the standard code sequence for PC relative loads/stores. It should be fine.

plutooo: Looks like the instruction encoding has a 9-bit immediate offset, this is not enough to cover it.

PR #41, opened 2018-01-31
Optimize for Cortex-A57.

profi200:


fincs:
If this gets merged: switch-examples need makefiles updated too.

PR #44, opened 2018-02-15
Add new domain handling commands, fix existing ones, other bugfixes.

TuxSH:
Also adds domain support to sm.h.

### DIFF
--- a/nx/include/switch/services/sm.h
+++ b/nx/include/switch/services/sm.h
@@ -12,14 +12,17 @@
 
 /// Service type.
 typedef enum {
-    ServiceType_Uninitialized, ///< Uninitialized service.
-    ServiceType_Normal,        ///< Normal service.
-    ServiceType_Override       ///< Service overriden in the homebrew environment.
+    ServiceType_Uninitialized,      ///< Uninitialized service.
+    ServiceType_Normal,             ///< Normal service.
+    ServiceType_Domain,             ///< Domain.
+    ServiceType_DomainSubservice,   ///< Domain subservice;
+    ServiceType_Override            ///< Service overriden in the homebrew environment.
 } ServiceType;
 
 /// Service object structure.
 typedef struct {
     Handle handle;
+    u32 object_id;
     ServiceType type;
 } Service;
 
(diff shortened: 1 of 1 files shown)


nx/include/switch/ipc.h
```diff
@@ -377,11 +379,11 @@ static inline Result ipcQueryPointerBufferSize(Handle session, size_t *size) {
         IpcParsedCommand r;
         ipcParse(&r);
 
-        struct ipcQueryPointerBufferSizeResponse {
+        struct {
             u64 magic;
             u64 result;
             u32 size;
-        } *raw = (struct ipcQueryPointerBufferSizeResponse*)r.Raw;
```
fincs: This breaks C++

nx/include/switch/ipc.h
```diff
@@ -420,11 +422,11 @@ static inline Result ipcConvertSessionToDomain(Handle session, u32* object_id_ou
         IpcParsedCommand r;
         ipcParse(&r);
 
-        struct ipcConvertSessionToDomainResponse {
+        struct {
             u64 magic;
             u64 result;
             u32 object_id;
-        } *raw = (struct ipcConvertSessionToDomainResponse*) r.Raw;
```
fincs: Same thing here

nx/include/switch/services/sm.h
```diff
@@ -68,9 +133,14 @@ static inline void serviceClose(Service* s) {
     switch (s->type) {
 
     case ServiceType_Normal:
+    case ServiceType_Domain:
```
fincs: What happens if you create a domain from an overriden service? It shouldn't use svcCloseHandle here.

TuxSH: You shouldn't convert an overriden service to a domain to begin with...

Do overriden domains exist/are they desirable?

fincs: FS comes to mind as a service that might be overriden; and that may benefit from domains iirc.

plutooo: Yeah we need to be able to override domains (but not subdomains)

PR #50, opened 2018-02-21
Add hidMouseRead, allows mouse position/velocity/scrollwheel velocity to be fetched

shinyquagsire23:
Thought I had this added but I guess not, only the mouse buttons were readable.


yellows8:
hidMouseRead() missing from .h.

PR #51, opened 2018-02-23
hid: Improve heuristics for P1_AUTO to switch between controllers, add CONTROLLER_ANY

shinyquagsire23:
[video of new heuristics](https://www.youtube.com/watch?v=1aiY7mYiDbM)

Heuristics used are the same as those in the home menu and other Nintendo system applets; the active joystick and active buttons are split. The controller used for buttons is the controller with newer inputs, and the controller used for joysticks is the first controller to be using the joystick. This way joystick input doesn't suddenly change, but the newest button inputs are always retrieved. Digital joystick values fall under the category of buttons and not joysticks.

In order to keep track of which inputs are newer, any buttonDown will store a timestamp of how many samples a controller has had active inputs (and the same goes for joystick values).

### DIFF
--- a/nx/include/switch/services/hid.h
+++ b/nx/include/switch/services/hid.h
@@ -310,7 +310,7 @@ typedef enum
     CONTROLLER_PLAYER_8 = 7,
     CONTROLLER_HANDHELD = 8,
     CONTROLLER_UNKNOWN  = 9,
-    CONTROLLER_P1_AUTO = 10, /// Not an actual HID-sysmodule ID. Only for hidKeys*(). Automatically uses CONTROLLER_PLAYER_1 when connected, otherwise uses CONTROLLER_HANDHELD.
+    CONTROLLER_P1_AUTO = 10, /// Not an actual HID-sysmodule ID. Only for hidKeys*(). Switches the commanding controller based on activity.
 } HidControllerID;
 
 typedef struct touchPosition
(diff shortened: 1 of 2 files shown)


plutooo:
Is this really how official sw works

nx/include/switch/services/hid.h
```diff
@@ -310,7 +310,8 @@ typedef enum
     CONTROLLER_PLAYER_8 = 7,
     CONTROLLER_HANDHELD = 8,
     CONTROLLER_UNKNOWN  = 9,
-    CONTROLLER_P1_AUTO = 10, /// Not an actual HID-sysmodule ID. Only for hidKeys*(). Automatically uses CONTROLLER_PLAYER_1 when connected, otherwise uses CONTROLLER_HANDHELD.
+    CONTROLLER_P1_AUTO = 10, /// Not an actual HID-sysmodule ID. Only for hidKeys*(). Switches the commanding controller between handheld and player 1 based on activity.
```
plutooo: Is there any reason not to "just do":

CONTROLLER_PLAYER_1 = BIT(0),
CONTROLLER_PLAYER_2 = BIT(1),
...
CONTROLLER_P1_AUTO = CONTROLLER_HANDHELD | CONTROLLER_PLAYER_1,
CONTROLLER_ANY = 0xFFFFFFFF

PR #55, opened 2018-02-25
switch to 16x16 console font

WinterMute:
16x16 font is way more readable on that screen


PoloNX:
Hello @WinterMute, how did you create this .bin ?

PR #57, opened 2018-02-27
mutex: Add TryLock() variants.

Armada651:
I haven't tested the implementation yet, so be careful about merging this. I'll test it soon.

### DIFF
--- a/nx/include/switch/kernel/mutex.h
+++ b/nx/include/switch/kernel/mutex.h
@@ -30,6 +30,13 @@ static inline void mutexInit(Mutex* m)
 void mutexLock(Mutex* m);
 
 /**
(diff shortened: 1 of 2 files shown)


nx/include/switch/kernel/mutex.h
```diff
@@ -29,6 +29,13 @@ static inline void mutexInit(Mutex* m)
  */
 void mutexLock(Mutex* m);
 
+/**
+ * @brief Attempts to lock a mutex without waiting.
+ * @param m Mutex object.
+ * @return 1 if the mutex has been acquired successfully, and 0 on contention.
+ */
+u32 mutexTryLock(Mutex* m);
```
plutooo: use bool?

PR #62, opened 2018-03-07
[hidvib] Rename vibration values variables

CTCaer:
Add max real value for amplitudes.
Amps > 1.0f are permitted and they are sent as 1.0f.

### DIFF
--- a/nx/include/switch/services/hid.h
+++ b/nx/include/switch/services/hid.h
@@ -541,7 +541,10 @@ static_assert(sizeof(HidSharedMemory) == 0x40000, "Hid Shared Memory structure h
 
 typedef struct HidVibrationValue
 {
-    float values[4];
+    float amp_low;   ///< Low Band amplitude. 1.0f: Max amplitude.
+    float freq_low;  ///< Low Band frequency in Hz.
+    float amp_high;  ///< High Band amplitude. 1.0f: Max amplitude.
+    float freq_high; ///< High Band frequency in Hz.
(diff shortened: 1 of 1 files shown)


tommai78101:
Try using:

`
typedef struct HidVibrationValue
{
    union {
      float values[4];
      struct {
        float amp_low;   ///< Low Band amplitude. 1.0f: Max amplitude.
        float freq_low;  ///< Low Band frequency in Hz.
        float amp_high;  ///< High Band amplitude. 1.0f: Max amplitude.
        float freq_high; ///< High Band frequency in Hz.
      }
    }
 } HidVibrationValue;
`

CTCaer:
Why? I think there's no need to use two different ways to apply our vibration pattern.

If it is about the failed check, this is not a problem. 
Because the PR I did in https://github.com/switchbrew/switch-examples/pull/21 needs to be merged first, for the PR here to build successfully.

tommai78101:
Hmmm.

I would've figured that having an array instead of typing out the variable names would be faster when coding.

CTCaer:
You can still set the values with an array 
example: 
```
HidVibrationValue vibvalues = {0.0f, 160.0f, 0.0f, 320.0f};
...
...
vibvalues = {1.0f, 261.7f, 0.9f, 328.0f};
```

tommai78101:
I see, but then if you have a union, doing that above gives the same result.

Hm... this is a doozy.

PR #63, opened 2018-03-12
Improve speed of framebuffer tiling

endrift:
This roughly doubles the performance of the copy in Yuzu, performance improvement not thoroughly tested on hardware yet. ~~I've also only thrown simple patterns at it so I'm not 100% sure it works properly yet.~~ Ok I threw a bitmap at it and it looks fine.

### DIFF
--- a/nx/source/gfx/gfx.c
+++ b/nx/source/gfx/gfx.c
@@ -563,19 +563,22 @@ void gfxFlushBuffers(void) {
 
         for (y=0; y<height; y++) {
             size_t row = y;
+            size_t y_base = y * width;
             if (g_gfx_drawflip) row = g_gfx_framebuf_display_height-1-row;
 
-            size_t y_offset = (row & 0x70) * 4 + (row & ~0x7F) * (g_gfx_framebuf_aligned_width / 16) * 4;
-            y_offset += (row & 8) * 4 + (row & 6) * 2 + (row & 1);
+            size_t y_offset = (row & 0x78) + (row & ~0x7F) * (g_gfx_framebuf_aligned_width / 16);
+            y_offset = y_offset * 4 + (row & 6) * 2 + (row & 1);
+
+            size_t x_offset = 0;
 
             for (x=0; x<width; x+=16) {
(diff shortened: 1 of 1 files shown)


hthh:
If you haven't, I'd recommend looking at https://fgiesen.wordpress.com/2011/01/17/texture-tiling-and-swizzling/ - it's a really interesting read (and the code is in the public domain).

yellows8:
@hthh This needs to be as fast (and hence simple) as possible, so no.

endrift:
Taking a look at this after sleeping I see an obvious way to remove a multiplication from the innermost loop, so I might want to do that today before this gets considered for merging.

hthh:
@yellows8 by all means don't use it if you want an easier-to-understand solution. But if it was unclear I only shared the post because it is a faster and simpler solution than the current code.

yellows8:
"it is a faster and simpler solution than the current code." Did you test this?

hthh:
No

yellows8:
The libt* blit func is slightly slower than current libnx impl, and that's based on the above anyway so...

hthh:
Yes, I wrote (based on the above) that blit code just to get the swizzle working (and figure out the pattern). It's never been optimized. I believe if you unroll it (as done in this PR, or in the example code linked from the post https://github.com/rygorous/vivante_swizzle/blob/master/main.cpp#L148 ) and simplify it, you could benefit from the lower loop overhead. Probably not a huge impact (most of the benefit is on the per-row code), just thought anyone working on optimizing this might be interested in reading about someone else's thorough efforts at optimizing tiling.

endrift:
The multiplication is optimized out anyway so it's probably not worth removing.

plutooo:
Did you try this with unconventional resolutions

endrift:
No, I was not aware you could put the device into unconventional resolutions.

@yellows8 said that testing this PR on actual hardware did not demonstrate a performance improvement anyway, so I might need to redo this PR. I'm assuming this means that optimizing for cache locality on reads instead of writes was a bad choice, but I'm not sure.

plutooo:
Okie, closing then

PR #65, opened 2018-03-15
add support for finding nxlink host

WinterMute:


nx/source/runtime/argv.c
```diff
@@ -136,6 +140,12 @@ void argvSetup(void)
         __system_argc++;
     }
 
+    if (strncmp(__system_argv[__system_argc - 1], "NXLINKED", 8) == 0 ) {
```
yellows8: Maybe check for __system_argc>=1?

plutooo: Maybe use __system_argv[0]

yellows8: But that will break using cwd at switch/.

PR #66, opened 2018-03-21
add socket flags translation

WinterMute:
This sorts out non blocking sockets.

### DIFF
--- a/nx/source/runtime/devices/socket.c
+++ b/nx/source/runtime/devices/socket.c
@@ -584,9 +584,35 @@ int ioctl(int fd, int request, ...) {
     }
 }
 
+#define O_NONBLOCK_NX 0x800
+
+#define ALL_NX   (O_NONBLOCK_NX)
+#define ALL_FLAGS (O_NONBLOCK)
+
+static int from_nx(int flags) {
+    int newflags = 0;
+
+    if(flags & O_NONBLOCK_NX)
+        newflags |= O_NONBLOCK;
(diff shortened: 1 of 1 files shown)


nx/source/runtime/devices/socket.c
```diff
@@ -584,9 +584,35 @@ int ioctl(int fd, int request, ...) {
     }
 }
 
+#define O_NONBLOCK_NX 0x800
+
+#define ALL_NX   (O_NONBLOCK_NX)
+#define ALL_FLAGS (O_NONBLOCK)
+
+static int from_nx(int flags) {
+    int newflags = 0;
```
plutooo: Where's O_NONBLOCK defined right now? 0x800 is what Linux uses I think, so this should already be correct? Can we fix this in the original header? If it's in newlib I guess it's not worth it to roll another version of it.

Follow codestyle so _FlagsFromNx()/_FlagsToNx()

yellows8: Current O_NONBLOCK value is wrong & is defined in newlib.

WinterMute: newlib uses 0x4000 - https://github.com/devkitPro/newlib/blob/master/newlib/libc/include/sys/_default_fcntl.h

Interestingly linux has `# define O_NONBLOCK      04000`

I did think about changing stuff around in newlib but do we know all the flags yet? We translated for 3DS in libctru.

PR #73, opened 2018-04-22
Implement svcCreatePort.

SciresM:


nx/source/kernel/svc.s
```diff
@@ -345,6 +345,17 @@ SVC_BEGIN svcWriteDebugProcessMemory
...
 SVC_END
 
+SVC_BEGIN svcCreatePort
+	stp x0, x1, [sp, #-16]!
+	svc 0x70
+	ldr x3, [sp]
+	str w1, [x3]
+	ldr x3, [sp, #8]
+	str w2, [x3]
+	add sp, sp, #16
```
fincs: I think this could be optimized to:
```asm
ldp x3, x4, [sp], #16
str w1, [x3]
str w2, [x4]
```

SciresM: It think it totally could be -- I had simply copied Nintendo's implementation. Is matching official implementation desirable, or should we fold in the optimization?

fincs: I had forgotten about this, but turns out I had previously done optimizations to the svc stubs (see cde9d05c8) and it was accepted, so why not do it here as well?

PR #74, opened 2018-04-26
Implement more SVCs

hexkyz:
Adding a few more syscalls


nx/include/switch/kernel/svc.h
```diff
@@ -819,4 +926,4 @@ Result svcStartProcess(Handle proc, s32 main_prio, s32 default_cpu, u32 stack_si
  */
 u64 svcCallSecureMonitor(SecmonArgs* regs);
 
-///@}
+///@}
```
fincs: Please don't delete trailing newline

nx/source/kernel/svc.s
```diff
@@ -430,4 +517,4 @@ SVC_BEGIN svcCallSecureMonitor
 	stp x4, x5, [x8, #0x20]
 	stp x6, x7, [x8, #0x30]
 	ret
-SVC_END
+SVC_END
```
fincs: Same here

PR #76, opened 2018-05-03
Add ResourceLimit SVCs, svcGetProcessInfo

SciresM:


nx/include/switch/kernel/svc.h
```diff
@@ -105,6 +105,32 @@ typedef enum {
     JitMapOperation_UnmapSlave=3, ///< Unmap slave.
 } JitMapOperation;
 
+/// Limitable Resources.
+typedef enum {
+    LimitableResource_Memory=0,           ///<How much memory can a process map?
```
yellows8: *.

PR #95, opened 2018-05-18
Add setGetSerialNumber

joel16:
Title. Let me know if I need to make any changes.

### DIFF
--- a/nx/include/switch/services/set.h
+++ b/nx/include/switch/services/set.h
@@ -66,3 +66,6 @@ void setsysExit(void);
 
 /// Gets the current system theme.
 Result setsysGetColorSetId(ColorSetId* out);
+
+/// Gets the serial number.
+Result setGetSerialNumber(char *serial);
--- a/nx/include/switch/services/set.h
+++ b/nx/include/switch/services/set.h
@@ -67,5 +67,8 @@ void setsysExit(void);
 /// Gets the current system theme.
 Result setsysGetColorSetId(ColorSetId* out);
 
-/// Gets the serial number.
-Result setGetSerialNumber(char *serial);
+/**
+ * @brief Gets the system's serial number.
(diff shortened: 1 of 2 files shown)


nx/include/switch/services/set.h
```diff
@@ -66,3 +66,6 @@ void setsysExit(void);
 
 /// Gets the current system theme.
 Result setsysGetColorSetId(ColorSetId* out);
+
+/// Gets the serial number.
```
fincs: Should document the parameters and mention that the output buffer needs to be able to hold at least 0x18 bytes.

yellows8:
setsys{name}

nx/include/switch/services/set.h
```diff
@@ -66,3 +66,9 @@ void setsysExit(void);
 
 /// Gets the current system theme.
 Result setsysGetColorSetId(ColorSetId* out);
+
+/**
+ * @brief Gets the system's serial number.
+ * @param serial Pointer to output the serial to. (The buffer size needs to be at least 0x18 bytes)
```
yellows8: *0x19

nx/source/services/set.c
```diff
@@ -346,3 +346,38 @@ Result setsysGetColorSetId(ColorSetId* out)
     return rc;
 
 }
+
+Result setsysGetSerialNumber(char *serial) {
+    IpcCommand c;
+    ipcInitialize(&c);
+
```
yellows8: Add: if (serial) memset(serial, 0, 0x19);

joel16: Should I change the declaration as well as the memcpy to 0x19 too?

yellows8: This is to make sure the string is NUL-terminated.

joel16: Yes I was aware of that, but I guess the result buffer should be fine then.

PR #96, opened 2018-05-19
Add string.h to remove compile error

joel16:
memset and memcpy requires string.h

### DIFF
--- a/nx/source/services/set.c
+++ b/nx/source/services/set.c
@@ -6,7 +6,6 @@
  * @copyright libnx Authors
  */
 #include <string.h>
-
 #include "types.h"
 #include "result.h"
 #include "arm/atomics.h"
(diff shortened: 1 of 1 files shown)


nx/source/services/set.c
```diff
@@ -5,6 +5,8 @@
  * @author yellows8
  * @copyright libnx Authors
  */
+#include <string.h>
+
```
yellows8: remove newline

joel16: I was just following the way it's used in acc.c -> https://github.com/switchbrew/libnx/blob/master/nx/source/services/acc.c#L1 but okay.

yellows8: Heh, no newline elsewhere :|

yellows8: Er nvm revert this, original didn't have newline & nothing(?) else has newline here either.

PR #103, opened 2018-05-27
Added accountGetUserCount and accountListAllUsers

3Daniel:


nx/include/switch/services/acc.h
```diff
@@ -33,6 +35,15 @@ Result accountInitialize(void);
 void accountExit(void);
 Service* accountGetService(void);
 
+/// Get the total number of user profiles
+Result accountGetUserCount(s32* user_count);
+
+/**
+ * @brief Get a list of all user IDs
+ * @param userIDs Pointer to array of user IDs, array must be exactly size 8
```
yellows8: Replace 8 with ACC_USER_LIST_SIZE.

Does official user-process sw have this size hard-coded to this?

yellows8:
Change it to ```static Result _accountListAllUsers```, and impl a wrapper func like this: ```Result _accountListAllUsers(u128* userIDs, size_t max_userIDs, size_t *actual_total);```

Where actual_total is calculated by how many userIDs are non-zero, stopping at the first all-zero one.

jyapayne:
@3Daniel Are you going to update this? I'm super looking forward to being able to use this in my homebrew, instead of asking the user to open a game first (which doesn't really work on hekate).

3Daniel:
@yellows8 can I please get a little clarification before I update this, just to make sure I understand you correctly.

I think it would be easier to use the `accountGetUserCount(s32* user_count)` function inside the wrapper to get the `actual_total` rather than calculate it.

I'm not clear on the purpose of `max_userIDs`. Is it so that the caller can specify a different maximum size for the `userIDs` array other than 8? ie. if `max_userIDs` is 3, then the `userIDs` array size is 3 and only the first 3 user IDs are returned?

yellows8:
accountGetUserCount is unnecessary, it would be just a simple for-loop. "Is it so that the caller can specify a different maximum size for the userIDs array other than 8?" Yes.

nx/source/services/acc.c
```diff
@@ -108,6 +109,38 @@ Result accountListAllUsers(u128* userIDs)
     return rc;
 }
 
+Result accountListAllUsers(u128* userIDs, size_t max_userIDs, size_t *actual_total)
+{
+    Result rc=0;
+
+    u128 *temp_userIDs;
+	temp_userIDs = malloc(sizeof(u128) * ACC_USER_LIST_SIZE);
```
fincs: Tab.

nx/source/services/acc.c
```diff
@@ -33,6 +34,113 @@ Service* accountGetService(void) {
...
+
+    rc = _accountListAllUsers(temp_userIDs);
+
+    if (R_SUCCEEDED(rc)) {
+        size_t total_userIDs = 0;
+
+        for (int i = 0; i < ACC_USER_LIST_SIZE; i++) {
+            if (temp_userIDs[i]) {
```
fincs: Should be the other way around, to match what yls8 requested: `if (!temp_userIDs[i]) break;`

nx/source/services/acc.c
```diff
@@ -33,6 +34,113 @@ Service* accountGetService(void) {
...
+
+        if (max_userIDs > total_userIDs) {
+            max_userIDs = total_userIDs;
+        }
+
+        for (int i = 0; i < max_userIDs; i++) {
+            userIDs[i] = temp_userIDs[i];
+        }
+
+        *actual_total = total_userIDs;
```
fincs: This is cumbersome. *actual_total should be assigned max_userIDs, which is the number of users that were returned.

3Daniel: Happy to change it, but this contradicts the original comment which said:
"Where actual_total is calculated by how many userIDs are non-zero, stopping at the first all-zero one."
The number which are non-zero is likely to be different to the max_userIDs.
Please confirm which way you'd prefer.

3Daniel: Yeah, that makes sense. I've changed it to be as you suggested now.

nx/source/services/acc.c
3Daniel: That is a lot cleaner, thank you for the suggestion :-)

nx/source/services/acc.c
3Daniel: Oops... that was silly of me, I think it should be fixed now.

nx/source/services/acc.c
fincs: The original statement is sort of ambiguous, but it's a common API convention to do it like what I said.

nx/include/switch/services/acc.h
```diff
@@ -33,6 +35,17 @@ Result accountInitialize(void);
 void accountExit(void);
 Service* accountGetService(void);
 
+/// Get the total number of user profiles
+Result accountGetUserCount(s32* user_count);
+
+/**
+ * @brief Get a list of all user IDs. The returned list will never be larger than ACC_USER_LIST_SIZE
```
yellows8: Periods missing.

nx/source/services/acc.c
```diff
@@ -33,6 +34,113 @@ Service* accountGetService(void) {
...
+    }
+
+    return rc;
+}
+
+Result accountListAllUsers(u128* userIDs, size_t max_userIDs, size_t *actual_total)
+{
+    Result rc=0;
+
+    u128 *temp_userIDs;
```
yellows8: Move to stack.

nx/source/services/acc.c
```diff
@@ -33,6 +34,113 @@ Service* accountGetService(void) {
...
+        for (total_userIDs = 0; total_userIDs < ACC_USER_LIST_SIZE; total_userIDs++) {
+            if (!temp_userIDs[total_userIDs])
+                break;
+        }
+
+        if (max_userIDs > total_userIDs) {
+            max_userIDs = total_userIDs;
+        }
+
+        for (int i = 0; i < max_userIDs; i++) {
```
yellows8: memcpy

nx/source/services/acc.c
```diff
@@ -112,8 +112,7 @@ static Result _accountListAllUsers(u128* userIDs)
 Result accountListAllUsers(u128* userIDs, size_t max_userIDs, size_t *actual_total)
 {
     Result rc=0;
-    u128 *temp_userIDs;
-    temp_userIDs = malloc(sizeof(u128) * ACC_USER_LIST_SIZE);
+    u128 *temp_userIDs = malloc(sizeof(u128) * ACC_USER_LIST_SIZE);
```
fincs: u128 temp_userIDs[ACC_USER_LIST_SIZE];

nx/source/services/acc.c
```diff
@@ -33,6 +34,108 @@ Service* accountGetService(void) {
...
+
+        if (max_userIDs > total_userIDs) {
+            max_userIDs = total_userIDs;
+        }
+
+        memcpy(userIDs, temp_userIDs, sizeof(u128) * max_userIDs);
+        *actual_total = max_userIDs;
+    }
+
+    free(temp_userIDs);
```
fincs: Remove this due to above being changed to stack.

fincs:
Any updates regarding the above comments? We'd like to get PRs merged in order to facilitate a release.

3Daniel:
Apologies for the delay. I have updated it to reflect your comments. I really appreciate you taking the time to provide feedback, even on small things. Thanks :-)

nx/source/services/acc.c
```diff
@@ -33,6 +34,106 @@ Service* accountGetService(void) {
...
+    }
+
+    return rc;
+}
+
+Result accountListAllUsers(u128* userIDs, size_t max_userIDs, size_t *actual_total)
+{
+    Result rc=0;
+    u128 temp_userIDs[ACC_USER_LIST_SIZE];
+
```
yellows8: memset(temp_userIDs, 0, sizeof(temp_userIDs));

fincs: Or just " = {0};` at the end of the declaration I guess.

nx/source/services/acc.c
```diff
@@ -1,4 +1,5 @@
 #include <string.h>
+#include <malloc.h>
```
yellows8: No longer needed.

PR #107, opened 2018-05-31
Added a few misc IPC wrappers

Adubbz:
Needed these for Gag Order, which is why they are not all for one service.


nx/source/services/set.c
```diff
@@ -348,6 +348,76 @@ Result setsysGetColorSetId(ColorSetId* out)
 
 }
 
+Result setsysGetSettingsItemValue(const char *name, const char *item_key, u64 *value_out) {
```
yellows8: Shouldn't value_out be an actual buffer with user-specified size?

nx/include/switch/services/set.h
```diff
@@ -108,3 +122,5 @@ Result setsysGetWirelessLanEnableFlag(bool *out);
  * @param out Pointer to output the status to. 
  */
 Result setsysGetBluetoothEnableFlag(bool *out);
+
```
yellows8: extra newlines

yellows8:
Can you add nsvm support for <v3.0 as well?

Adubbz:
ns:vm didn't actually exist until 3.0.0, so I'm afraid not.

yellows8:
As in use ns:am serv instead of ns:vm, for cmd 1200.

Adubbz:
Ah okay, yeah I can do that

yellows8:
Update nsvmInitialize to return 0 for <v3.0, likewise for nsvmExit(just return). Update nsvmGetSafeSystemVersion to return the incompat-sysver error for <v3.0.

nx/source/services/set.c
```diff
@@ -348,6 +348,76 @@ Result setsysGetColorSetId(ColorSetId* out)
 
 }
 
+Result setsysGetSettingsItemValue(const char *name, const char *item_key, void *value_out, size_t value_out_size) {
+    IpcCommand c;
+    ipcInitialize(&c);
+    ipcAddSendStatic(&c, name, SET_MAX_NAME_SIZE, 0);
```
yellows8: Does official sw use fixed SET_MAX_NAME_SIZE or the actual stringlen?

WinterMute:
I'm not sure that "GetSettings" adds any information to the functions being added here. Would setsysItemValue and setsysItemValueSize be better?

nx/source/services/set.c
Adubbz: Official software does use this, however it was meant to be 0x48 not 0x30, since it seems the Pegaswitch script I was referencing was wrong.

Adubbz:
I mainly added GetSettings for consistency with the names on the wiki, but it can be changed if that's desirable

yellows8:
@WinterMute Those are official names.

yellows8:
Can you copy the input strings into stack buffers which actually have SET_MAX_NAME_SIZE?(memset+strncpy) Official sw does this.

nx/include/switch/services/pm.h
```diff
@@ -27,3 +27,4 @@ Result pmdmntEnableDebugForApplication(Handle* handle_out);
 Result pminfoGetTitleId(u64* title_id_out, u64 pid);
 
 Result pmshellLaunchProcess(u32 launch_flags, u64 titleID, u64 storageID, u64 *pid);
+Result pmshellTerminateProcessByTitleId(u64 title_id);
```
yellows8: Use same titleID name as prev line.

nx/include/switch/services/set.h
```diff
@@ -107,4 +121,4 @@ Result setsysGetWirelessLanEnableFlag(bool *out);
  * @brief Gets the bluetooth status.
  * @param out Pointer to output the status to. 
  */
-Result setsysGetBluetoothEnableFlag(bool *out);
+Result setsysGetBluetoothEnableFlag(bool *out);
```
yellows8: Revert to the original.

nx/source/services/ns.c
```diff
@@ -116,3 +116,103 @@ Result nsGetApplicationControlData(u8 flag, u64 titleID, NsApplicationControlDat
...
+
+Result nsvmInitialize(void)
+{
+    if (!kernelAbove300())
+        return 0;
+
+    atomicIncrement64(&g_nsvmRefCnt);
+
+    if (serviceIsActive(&g_nsvmSrv))
+        return MAKERESULT(Module_Libnx, LibnxError_AlreadyInitialized);
```
yellows8: return 0;

nx/include/switch/services/ns.h
```diff
@@ -17,3 +17,9 @@ Result nsInitialize(void);
 void nsExit(void);
 
 Result nsGetApplicationControlData(u8 flag, u64 titleID, NsApplicationControlData* buffer, size_t size, size_t* actual_size);
+
+Result nsvmInitialize(void);
+void nsvmExit(void);
+
+Result nsvmNeedsUpdateVulnerability(u8 *out);
```
fincs: Should have been `bool`

nx/include/switch/services/set.h
```diff
@@ -67,6 +69,18 @@ void setsysExit(void);
...
 Result setsysGetColorSetId(ColorSetId* out);
 
+/**
+ * @brief Gets the size of a settings item value.
+ * @param out Pointer to output the size to.
+ */
+Result setsysGetSettingsItemValueSize(const char *name, const char *item_key, u64 *size_out);
+
+/**
+ * @brief Gets the value of a settings item value.
```
yellows8: Extra 'value' at the end.

PR #112, opened 2018-06-11
Improve IpcParsedCommand domain message info.

SciresM:
Currently it's hard to use IpcParsedCommand for domain messages.

This fixes that.

### DIFF
--- a/nx/include/switch/kernel/ipc.h
+++ b/nx/include/switch/kernel/ipc.h
@@ -636,12 +636,12 @@ static inline Result ipcParseForDomain(IpcParsedCommand* r) {
         case DomainMessageType_Close:
             break;
         default:
-            return 0xF601;
+            return 0x1A60A;
     }
     r->ThisObjectId = hdr->ThisObjectId;
     r->NumObjectIds = hdr->NumObjectIds > 8 ? 8 : hdr->NumObjectIds;
     if ((uintptr_t)object_ids + sizeof(u32) * r->NumObjectIds - (uintptr_t)armGetTls() >= 0x100) {
-        return 0xF601;
+        return 0x1D60A;
     }
     for(size_t i = 0; i < r->NumObjectIds; i++)
         r->ObjectIds[i] = object_ids[i];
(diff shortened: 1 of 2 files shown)


nx/include/switch/kernel/ipc.h
```diff
@@ -45,9 +45,24 @@ typedef enum {
...
     IpcCommandType_RequestWithContext = 6,
-    IpcCommandType_ControlWithContext = 7
+    IpcCommandType_ControlWithContext = 7,
 } IpcCommandType;
 
+typedef enum {
+    DomainMessageType_Invalid = 0,
+    DomainMessageType_SendMessage = 1,
+    DomainMessageType_Close = 2,
+} DomainMessageType;
```
yellows8: Ipc prefix missing + elsewhere.

SciresM: I was matching the existing style -- @plutooo called the domain header type "DomainMessageHeader", does that also need to change?

nx/include/switch/kernel/ipc.h
```diff
@@ -618,8 +629,20 @@ static inline Result ipcParseForDomain(IpcParsedCommand* r) {
...
     r->Raw = (void*)(((uintptr_t) r->Raw) + sizeof(DomainMessageHeader));
 
+    r->IsDomainMessage = true;
+    r->MessageType = (DomainMessageType)(hdr->Type);
+    switch (r->MessageType) {
+        case DomainMessageType_SendMessage:
+        case DomainMessageType_Close:
+            break;
+        default:
+            return 0xF601;
```
yellows8: Does official sw return this?

nx/include/switch/kernel/ipc.h
```diff
@@ -636,12 +636,12 @@ static inline Result ipcParseForDomain(IpcParsedCommand* r) {
         case DomainMessageType_Close:
             break;
         default:
-            return 0xF601;
+            return 0x1A60A;
```
yellows8: Change these to a new libnx error.

fincs: I think this is supposed to match official sw, not sure.

PR #118, opened 2018-06-14
translate horizon errno to newlib errno

WinterMute:
Should fix curl/mbedtls issue


y4my4my4m:
I've tested it from a clean install. Works as intended.
Fantastic, thank you for looking into the issue!

nx/source/runtime/devices/convert_errno.c
```diff
@@ -0,0 +1,165 @@
...
+    ERFKILL         ,
+    EHWPOISON       ,
+};
+
+#define _UNKNOWN_ERROR_OFFSET	10000
+
+int _convert_errno(int bsdErrno)
+{
+
+    if (bsdErrno < 0 || bsdErrno > sizeof(_error_code_map) || _error_code_map[bsdErrno] == 0)
```
yellows8: ```>=```

PR #122, opened 2018-06-25
ns:dev: Add TerminateProcess and TerminateProgram

SciresM:


nx/source/services/ns.c
```diff
@@ -289,4 +303,71 @@ Result nsvmGetSafeSystemVersion(u16 *out)
     }
 
     return rc;
-}
\ No newline at end of file
+}
+
+
```
yellows8: Extra newline.

PR #124, opened 2018-06-25
allow user additions to init/exit code

WinterMute:


nx/source/runtime/init.c
```diff
@@ -114,10 +116,15 @@ void __attribute__((weak)) __appInit(void)
         fatalSimple(MAKERESULT(Module_Libnx, LibnxError_InitFail_FS));
 
     fsdevMountSdmc();
+
+    if(&userAppInit) userAppInit();
```
yellows8: Missing space between if and (, + below.

PR #125, opened 2018-06-25
Add Semaphore

Kevoot:
Implement basic semaphore, no sem_post (yet)

### DIFF
--- a/nx/include/switch/kernel/semaphore.h
+++ b/nx/include/switch/kernel/semaphore.h
@@ -1,29 +1,24 @@
-// Copyright 2018 Kevoot
-
-#ifndef __SEMAPHORE_H
-#define __SEMAPHORE_H
+/**
+ * @file semaphore.h
+ * @brief Thread synchronization based on Mutex.
+ * @author Kevoot
+ * @copyright libnx Authors
+ */
+#pragma once
 
 #include "mutex.h"
 
-#define SEM_DOWN false
-#define SEM_UP   true
-
-#define EBUSY 1
-
-typedef struct sem_t
+typedef struct Semaphore
 {
-    bool flag;
+    bool is_up;
     Mutex mutex;
-} sem_t;
+} Semaphore;
 
-void sem_init(sem_t *);
-void sem_uninit(sem_t *);
+void semaphoreInit(Semaphore *);
 /* TODO */
-void set_post(sem_t *);
-void sem_up(sem_t *);
-void sem_down(sem_t *);
-void sem_wait(sem_t *);
-void sem_waitup(sem_t *);
-bool sem_isup(sem_t *);
-
-#endif
+void semaphorePost(Semaphore *);
+void semaphoreUp(Semaphore *);
+void semaphoreDown(Semaphore *);
+void semaphoreWait(Semaphore *);
+void semaphoreWaitUp(Semaphore *);
+bool semaphoreIsUp(Semaphore *);
--- a/nx/include/switch/kernel/semaphore.h
+++ b/nx/include/switch/kernel/semaphore.h
@@ -1,7 +1,7 @@
 /**
  * @file semaphore.h
  * @brief Thread synchronization based on Mutex.
- * @author Kevoot
+ * @author SciresM & Kevoot
  * @copyright libnx Authors
  */
 #pragma once
@@ -9,14 +9,36 @@
 #include "mutex.h"
 #include "condvar.h"
 
+/// Semaphore structure.
 typedef struct Semaphore
 {
-    CondVar condvar;
-    Mutex mutex;
-    u64 count;
+    CondVar condvar; ///< Conditional Variable Object.
+    Mutex   mutex;   ///< Mutex Object.
+    u64     count;   ///< Internal Counter.
 } Semaphore;
 
-void semaphoreInit(Semaphore *, u64);
-void semaphoreSignal(Semaphore *);
-void semaphoreWait(Semaphore *);
-bool semaphoreTryWait(Semaphore *);
+/**
+ * @brief Initializes a semaphore and its internal counter.
+ * @param s Semaphore object.
+ * @param initial_count initial value for internal counter (typically the # of free resources).
+ */
+void semaphoreInit(Semaphore * s, u64 initial_count);
+
+/**
+ * @brief Increments the Semaphore to allow other threads to continue.
+ * @param s Semaphore object.
+ */
+void semaphoreSignal(Semaphore * s);
+
+/**
+ * @brief Decrements Semaphore and waits if 0.
+ * @param s Semaphore object.
+ */
+void semaphoreWait(Semaphore * s);
+
+/**
+ * @brief Attempts to get lock without waiting.
+ * @param s Semaphore object.
+ * @return 1 if no wait is needed for lock, 0 otherwise.
+ */
+bool semaphoreTryWait(Semaphore * s);
--- a/nx/include/switch/kernel/semaphore.h
+++ b/nx/include/switch/kernel/semaphore.h
@@ -39,6 +39,6 @@ void semaphoreWait(Semaphore * s);
 /**
  * @brief Attempts to get lock without waiting.
  * @param s Semaphore object.
- * @return 1 if no wait is needed for lock, 0 otherwise.
+ * @return true if no wait and successful lock, false otherwise.
  */
 bool semaphoreTryWait(Semaphore * s);
--- a/nx/include/switch/kernel/semaphore.h
+++ b/nx/include/switch/kernel/semaphore.h
@@ -22,23 +22,23 @@ typedef struct Semaphore
  * @param s Semaphore object.
  * @param initial_count initial value for internal counter (typically the # of free resources).
  */
-void semaphoreInit(Semaphore * s, u64 initial_count);
+void semaphoreInit(Semaphore *s, u64 initial_count);
 
 /**
  * @brief Increments the Semaphore to allow other threads to continue.
  * @param s Semaphore object.
  */
-void semaphoreSignal(Semaphore * s);
+void semaphoreSignal(Semaphore *s);
 
 /**
  * @brief Decrements Semaphore and waits if 0.
  * @param s Semaphore object.
  */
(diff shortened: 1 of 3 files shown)


nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,29 @@
+// Copyright 2018 Kevoot
+
+#ifndef __SEMAPHORE_H
```
fincs: Use #pragma once instead of include guard.

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,29 @@
+// Copyright 2018 Kevoot
```
fincs: Use Doxygen file header.

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,29 @@
...
+
+#ifndef __SEMAPHORE_H
+#define __SEMAPHORE_H
+
+#include "mutex.h"
+
+#define SEM_DOWN false
+#define SEM_UP   true
+
+#define EBUSY 1
```
fincs: Do not redefine this, I don't see why you need to use EBUSY, it's not used anywhere.

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,29 @@
...
+
+typedef struct sem_t
```
fincs: According to naming convention this should be `Semaphore`

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,29 @@
...
+{
+    bool flag;
+    Mutex mutex;
+} sem_t;
+
+void sem_init(sem_t *);
```
fincs: According to naming convention this should be `semaphoreInit` (same for all the other ones)

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,29 @@
+// Copyright 2018 Kevoot
+
+#ifndef __SEMAPHORE_H
+#define __SEMAPHORE_H
+
+#include "mutex.h"
+
+#define SEM_DOWN false
```
fincs: I don't think these defines are necessary (see below comment in the struct).

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,29 @@
...
+#define SEM_UP   true
+
+#define EBUSY 1
+
+typedef struct sem_t
+{
+    bool flag;
```
fincs: Could be renamed to `is_up`.

nx/source/kernel/semaphore.c
```diff
@@ -0,0 +1,58 @@
...
+    while (true) {
+        mutexLock(&sem->mutex);
+        
+        if (sem->flag == flag) {
+            mutexUnlock(&sem->mutex);
+            break;
+        }
+
+        mutexUnlock(&sem->mutex);
+        svcSleepThread(1000000);
```
fincs: This doesn't seem right to me. Isn't there a better way to do this without polling?

nx/source/kernel/semaphore.c
```diff
@@ -0,0 +1,58 @@
...
+    }
+}
```
fincs: Missing newline at end of file.

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,29 @@
...
+void sem_uninit(sem_t *);
+/* TODO */
+void set_post(sem_t *);
+void sem_up(sem_t *);
+void sem_down(sem_t *);
+void sem_wait(sem_t *);
+void sem_waitup(sem_t *);
+bool sem_isup(sem_t *);
+
+#endif
```
fincs: Missing newline at end of file.

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,29 @@
...
+#define EBUSY 1
+
+typedef struct sem_t
+{
+    bool flag;
+    Mutex mutex;
+} sem_t;
+
+void sem_init(sem_t *);
+void sem_uninit(sem_t *);
```
fincs: Remove this function, it's not needed.

nx/source/kernel/semaphore.c
Kevoot: Pretty standard way of handling semaphores is to sleep and check. I can reduce the time down but don't see any benefit in removing it.

nx/include/switch/kernel/semaphore.h
Kevoot: Do we really want to deviate that far from the posix standard names? Just more time on people writing ports if there always has to be a wrapper around everything just because of naming conventions. (Same as below)

fincs: libnx synchronization primitives aren't POSIX based... Look at the other ones...

nx/source/kernel/semaphore.c
Kevoot: Not sure why the review didn't pick this up. It's been updated accordingly

SciresM:
@Kevoot this isn't the proper way to implement a semaphore in Horizon OS -- a semaphore should never be polling, it should block when waiting and be awoken by the kernel when it's able to acquire the semaphore.

The way to do this is via a CondVar + a Mutex -- I implemented it in as a C++ object in  [libstratosphere](https://github.com/Atmosphere-NX/Atmosphere/blob/master/stratosphere/libstratosphere/include/stratosphere/hossynch.hpp#L77) for Atmosphere before. If you want to correct your implementation to be non-polling go ahead, otherwise I can do a quick implementation.

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,22 @@
+/**
+ * @file semaphore.h
+ * @brief Thread synchronization based on Mutex.
+ * @author Kevoot
```
fincs: I guess SciresM's name should be added here now.

nx/source/kernel/semaphore.c
```diff
@@ -0,0 +1,35 @@
...
+void semaphoreSignal(Semaphore *sem) {
+    mutexLock(&sem->mutex);
+    sem->count++;
+    condvarWakeOne(&sem->condvar);
+    mutexUnlock(&sem->mutex);
+}
+
+void semaphoreWait(Semaphore *sem) {
+    mutexLock(&sem->mutex);
+    while(!(&sem->count)) {
```
fincs: Should be `while (!sem->count)` (note the space too)

nx/source/kernel/semaphore.c
```diff
@@ -0,0 +1,35 @@
...
+        condvarWait(&sem->condvar);
+    }
+    sem->count--;
+    mutexUnlock(&sem->mutex);
+}
+
+bool semaphoreTryWait(Semaphore *sem) {
+    mutexLock(&sem->mutex);
+    bool success = false;
+    if(sem->count) {
```
fincs: Space after `if`

nx/source/kernel/semaphore.c
```diff
@@ -0,0 +1,35 @@
...
+        (sem->count)--;
```
fincs: Redundant parentheses.

nx/source/kernel/semaphore.c
```diff
@@ -0,0 +1,35 @@
...
+        success = true;
+    }
+    return success;
+}
```
fincs: Missing newline at eof.

fincs:
Also, to fix CI build, please rebase on top of latest master.

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,22 @@
...
+#include "condvar.h"
+
+typedef struct Semaphore
+{
+    CondVar condvar;
+    Mutex mutex;
+    u64 count;
+} Semaphore;
+
+void semaphoreInit(Semaphore *, u64);
```
fincs: I think this should have Doxygen documentation too, like the other synchronization primitives. Also, the parameter names are missing in the prototypes. `Semaphore*` should probably be `s` and the `u64` should be `initial_count`.

fincs:
... I'm not sure what you did, but that wasn't a rebase...

You're supposed to do the rebase by pulling from origin, and then doing git rebase master on your branch. Currently it looks like this for me:
image

WinterMute:
This should fix it

```
git checkout semaphore
git reset --hard e8fc3dbe
git checkout master
git remote add upstream git@github.com:switchbrew/libnx.git
git fetch upstream
git merge upstream/master
git checkout semaphore
git rebase master
git push -f
```

fincs:
(and also reapply the changes I requested that were, ahem, "squashed" with the "rebase")

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * @brief Decrements Semaphore and waits if 0.
+ * @param s Semaphore object.
+ */
+void semaphoreWait(Semaphore * s);
+
+/**
+ * @brief Attempts to get lock without waiting.
+ * @param s Semaphore object.
+ * @return 1 if no wait is needed for lock, 0 otherwise.
```
fincs: true/false, not 1/0

Kevoot: Copied style from mutex.h, assumed this was standard as well.

nx/include/switch/kernel/semaphore.h
```diff
@@ -0,0 +1,44 @@
...
+    Mutex   mutex;   ///< Mutex Object.
+    u64     count;   ///< Internal Counter.
+} Semaphore;
+
+/**
+ * @brief Initializes a semaphore and its internal counter.
+ * @param s Semaphore object.
+ * @param initial_count initial value for internal counter (typically the # of free resources).
+ */
+void semaphoreInit(Semaphore * s, u64 initial_count);
```
yellows8: Change these "Semaphore * s" to match the .c.

PR #127, opened 2018-06-26
Fatal: Only exit process if required, exit cleanly.

SciresM:


nx/include/switch/services/fatal.h
```diff
@@ -26,4 +26,4 @@ void NORETURN fatalSimple(Result err);
  * @param err[in] Type of fatal error to throw.
  * @note This function does not return.
```
yellows8: outdated comment

PR #132, opened 2018-07-15
adjust fsdev_fixpath for cwd changes

WinterMute:


nx/source/runtime/devices/fs_dev.c
```diff
@@ -199,6 +199,13 @@ fsdev_fixpath(struct _reent *r,
     strncpy(__fixedpath, path, PATH_MAX);
   else
   {
+    size_t __cwd_len = strlen(__cwd);
```
yellows8: Shouldn't this be in fsdev chdir, since that's where __cwd is set?

PR #135, opened 2018-07-21
Fixed hid in sysmodules

jakibaki:


yellows8:
Sysmodules are supposed to disable hid init/exit. https://github.com/switchbrew/libnx/blob/master/nx/source/runtime/init.c#L89

jakibaki:
Without this pr hidInitialize would fail if called from a sysmodule so the sysmodule couldn't read the (for example) joycon inputs.

This fixes that and allows the joycon inputs to be read as a sysmodule.

nx/source/services/hid.c
```diff
@@ -49,8 +49,9 @@ Result hidInitialize(void)
     u64 AppletResourceUserId;
 
     rc = appletGetAppletResourceUserId(&AppletResourceUserId);
+    // If this failed (for example because we're a sysmodule) AppletResourceUserId stays zero
     if (R_FAILED(rc))
-        return rc;
+        AppletResourceUserId = 0;
```
yellows8: Does hidSetNpadJoyAssignmentModeDual() work with this same change if the below check is removed?

jakibaki: Yeah, it does!

I'll update my pr with better changes.

nx/source/services/hid.c
```diff
@@ -569,8 +570,7 @@ Result hidMergeSingleJoyAsDualJoy(HidControllerID id0, HidControllerID id1) {
 
     rc = appletGetAppletResourceUserId(&AppletResourceUserId);
     if (R_FAILED(rc))
-        return rc;
-
+        AppletResourceUserId = 0;
```
fincs: The newline was mistakenly removed here.

jakibaki: you're right, I'll fix that.

PR #140, opened 2018-07-27
Wrapped and implemented fsOpenBisFilesystem()

klockee:
Implemented fsOpenBisFilesystem()

### DIFF
--- a/nx/source/services/fs.c
+++ b/nx/source/services/fs.c
@@ -104,41 +104,41 @@ Result fsOpenBisStorage(FsStorage* out, u32 PartitionId) {
 Result fsOpenBisFileSystem(FsFileSystem* out, u32 PartitionId, const char* string) {
     IpcCommand c;
     ipcInitialize(&c);
-    
+
     char tmpstr[FS_MAX_PATH] = {0};
     strncpy(tmpstr, string, sizeof(tmpstr)-1);
     ipcAddSendStatic(&c, tmpstr, sizeof(tmpstr), 0);
-   
+
     struct {
         u64 magic;
         u64 cmd_id;
         u32 PartitionId;
     } *raw;
- 
+
     raw = ipcPrepareHeader(&c, sizeof(*raw));
- 
+
     raw->magic = SFCI_MAGIC;
     raw->cmd_id = 11;
     raw->PartitionId = PartitionId;
- 
+
     Result rc = serviceIpcDispatch(&g_fsSrv);
- 
+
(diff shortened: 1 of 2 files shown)


nx/source/services/fs.c
```diff
@@ -101,6 +101,43 @@ Result fsOpenBisStorage(FsStorage* out, u32 PartitionId) {
     return rc;
 }
 
+Result fsOpenBisFileSystem(FsFileSystem* out, u32 PartitionId, const char* string) {
+    IpcCommand c;
+    ipcInitialize(&c);
+    ipcAddSendStatic(&c, string, FS_MAX_PATH, 0);
+    struct {
```
yellows8: add newline

rajkosto:
If youre taking in a string pointer, either take also a length or do a strlen(str)+1 in the buffer send, to avoid sending garbage over IPC (this function doesnt really need to take the string pointer though, since official software always just passes 1 zero byte to the IPC)

nx/source/services/fs.c
```diff
@@ -101,6 +101,44 @@ Result fsOpenBisStorage(FsStorage* out, u32 PartitionId) {
     return rc;
 }
 
+Result fsOpenBisFileSystem(FsFileSystem* out, u32 PartitionId, const char* string) {
+    IpcCommand c;
+    ipcInitialize(&c);
```
yellows8: Add: char tmpstr[FS_MAX_PATH] = {0};

nx/source/services/fs.c
```diff
@@ -101,6 +101,44 @@ Result fsOpenBisStorage(FsStorage* out, u32 PartitionId) {
...
+    ipcAddSendStatic(&c, string, strlen(string) + 1, 0);
```
yellows8: strncpy(tmpstr, string, sizeof(tmpstr)-1);

Then ipcAddSendStatic with tmpstr and sizeof(tmpstr).

nx/source/services/fs.c
```diff
@@ -104,38 +104,41 @@ Result fsOpenBisStorage(FsStorage* out, u32 PartitionId) {
...
+    strncpy(tmpstr, string, sizeof(tmpstr)-1);
+    ipcAddSendStatic(&c, tmpstr, sizeof(tmpstr), 0);
+   
     struct {
         u64 magic;
         u64 cmd_id;
         u32 PartitionId;
     } *raw;
-
+
```
fincs: Revert these whitespace changes

PR #145, opened 2018-07-31
Add bpc service with ShutdownSystem and RebootSystem commands (1.0.0 firm fix)

XorTroll:
This time "bpc:c" is the service's name if firmware is not higher than 1.0.0


nx/source/services/bpc.c
```diff
@@ -0,0 +1,86 @@
+#include "types.h"
+#include "result.h"
+#include "arm/atomics.h"
+#include "kernel/ipc.h"
```
fincs: Missing `#include "kernel/detect.h"` (see Doozer error)

PR #147, opened 2018-08-04
home button block/unblock via hid:dbg and hid:sys

p-sam:
Adds the ability to disable the home button, and to enable it back. It uses 2 differents services, the disable method being in hid:dbg, and the enable one being in hid:sys.

A quick example using the 2 functions can be found here at https://github.com/p-sam/nx-hmblock. 

The `hidDbgDeactivateHomeButton` ipc calls exibits a weird behavior: Result being 0x16eca on actual success. When returning 0, the home button is still usable, and won't be disabled until a subsequent call actually returns this value 0x16eca. Calling it while the button is already disabled also returns 0x16eca. I've made the choice to loop until either a count of tries exceed 100 or we get rc != 0, returning 0 if rc was 0x16eca, i'm open to suggestions if that can be improved.


fincs:
Sounds like `hid:dbg` is an internal debugging service, which could change at any time. Are you sure that official applications block the HOME button this way?

p-sam:
No, I believe official apps use appletOE

fincs:
Could you make libnx use the same mechanisms that official apps use to block the HOME button?

p-sam:
I believe this is another set of change that would go in another PR, feel free to decline this PR if you don't want those fonctions until then.

fincs:
Ok then.

PR #153, opened 2018-08-16
Fix fsOpenFileSystem

rajkosto:
Actually works now instead of always returning port remote dead
Thanks @misson20000 for finding out the switchbrew parameters for it were wrong

### DIFF
--- a/nx/include/switch/services/fs.h
+++ b/nx/include/switch/services/fs.h
@@ -179,8 +179,8 @@ typedef enum
 } FsFileSystemType;
 
 /// Mount requested filesystem type from content file
-Result fsOpenFileSystem(FsFileSystem* out, FsFileSystemType fsType, const char* contentPath); /// only on 1.0.0
-Result fsOpenFileSystemWithId(FsFileSystem* out, u64 titleId, FsFileSystemType fsType, const char* contentPath); /// 2.0.0+
+Result fsOpenFileSystem(FsFileSystem* out, FsFileSystemType fsType, const char* contentPath);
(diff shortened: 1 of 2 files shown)


rajkosto:
Maybe the 2 functions for 1.0.0 and 2.0.0+ should be merged into one, as well ?

yellows8:
If the args are wrong update wiki.

rajkosto:
Done

yellows8:
"merged into one" But the args are different...

rajkosto:
fsOpenFileSystemWithId can just ignore the titleid passed in on 1.0.0. You can often leave it at 0 or -1 on 2.0.0+ anyway.

rajkosto:
This way is probably way more developer-friendly.

nx/include/switch/services/fs.h
```diff
@@ -179,8 +179,8 @@ typedef enum
 } FsFileSystemType;
 
 /// Mount requested filesystem type from content file
-Result fsOpenFileSystem(FsFileSystem* out, u64 titleId, FsFileSystemType fsType); /// only on 1.0.0, only works with registered content
-Result fsOpenFileSystemWithId(FsFileSystem* out, u64 titleId, FsFileSystemType fsType, const char* contentPath); /// 2.0.0+, contentPath must be resolved manually
+Result fsOpenFileSystem(FsFileSystem* out, FsFileSystemType fsType, const char* contentPath);
```
yellows8: Should add a note about sys-versions here, since either cmd can be used now.

PR #165, opened 2018-09-06
Add appletBeginBlockingHomeButton() and appletEndBlockingHomeButton()

neonsea:


XorTroll:
AppletType can also be SystemApplication, I tested it before.

neonsea:
Thanks for the heads up. I was unsure of your indentation style for long lines, so I left it as a half-indent - let me know if that's okay.

nx/include/switch/services/applet.h
```diff
@@ -53,6 +53,10 @@ Result appletCreateManagedDisplayLayer(u64 *out);
 
 Result appletGetDesiredLanguage(u64 *LanguageCode);
 
+Result appletBeginBlockingHomeButton(s64 val);
```
fincs: What's this "s64 val"? Is it a bool, is it an enum?

XorTroll:
SWIPC has no documentation about it. but retail apps use 0, and on my tests I used 0 and it worked successfully.

neonsea:
I just tested, and 0 seems to provide the intended behavior, and if this is what official apps use, I think this should be hardcoded to 0. Also added a notice indicating that this only works in regularapps.

yellows8:
The official wrapper funcs calling "BeginBlockingHomeButtonShortAndLongPressed" and "BeginBlockingHomeButton" hard-code the input to 0, however the latter has an additional separate func which passes the input param to the {cmd func}.

For the cmd impl'd here, the input is nanoseconds. Revert the hard-coded input change and update the comment.

PR #169, opened 2018-09-14
Resolve trivial -Wmissing-prototypes warnings

lioncash:
Gets rid of trivial warnings caused by headers not being included into the relevant .c file, empty parameter lists missing void, or functions missing a prototype in the header.

I've included the Nvidia ioctl prototypes that were missing, but can omit those if that interface is still in flux.


fincs:
Ignore the doozer error, that's just example building being broken due to a breaking change. This looks good to me.

PR #170, opened 2018-09-15
Add psm:GetChargerType

endrift:
Don't know if the names are correct but the only thing that returns ChargerType 1 is the official charging cable (be it directly or through the dock)--every other USB-C device I've plugged in returns 2.

### DIFF
--- a/nx/include/switch/services/psm.h
+++ b/nx/include/switch/services/psm.h
@@ -7,7 +7,14 @@
 #pragma once
 #include "../types.h"
 
+typedef enum {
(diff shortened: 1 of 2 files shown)


nx/include/switch/services/psm.h
```diff
@@ -7,7 +7,14 @@
 #pragma once
 #include "../types.h"
 
+typedef enum {
+    ChargerType_None = 0,
+    ChargerType_Charger = 1,
```
yellows8: Add comments for these.

PR #173, opened 2018-09-18
applet: Add appletSelfExit()

leo60228:
Add the [Exit](http://switchbrew.org/index.php/AM_services#Exit) command to applet. This enables exiting an application back to the home menu. It was named `appletSelfExit()` instead of `appletExit()` since `appletExit()` already existed.

### DIFF
--- a/nx/include/switch/services/applet.h
+++ b/nx/include/switch/services/applet.h
@@ -72,6 +72,12 @@ Result appletSetScreenShotPermission(s32 val);
 Result appletSetScreenShotImageOrientation(s32 val);
 
 /**
+ * @brief Exits the application.
+ * @note Can only be used with AppletType_Application.
+ */
+Result appletSelfExit(void);
+
+/**
  * @brief Processes the current applet status. Generally used within a main loop.
  * @return Whether the application should continue running.
  */
--- a/nx/source/services/applet.c
+++ b/nx/source/services/applet.c
@@ -1071,6 +1071,37 @@ Result appletSetScreenShotImageOrientation(s32 val) {
     return rc;
 }
 
+Result appletSelfExit(void) {
+    IpcCommand c;
+    ipcInitialize(&c);
+
+    struct {
+        u64 magic;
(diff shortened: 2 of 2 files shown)


nx/include/switch/services/applet.h
```diff
@@ -71,6 +71,11 @@ Result appletSetScreenShotPermission(s32 val);
 
 Result appletSetScreenShotImageOrientation(s32 val);
 
+/**
+ * @brief Exits the application.
```
yellows8: *the current applet

(Assuming AppletType doesn't matter?)

yellows8: What AppletType(s) was this tested with?

leo60228: I've only tested it with an Application. Switchbrew doesn't say either way, I'll check SwIPC.

leo60228: SwIPC doesn't say either. I don't know how to build an Applet, though.

leo60228: Actually, I'll see what it does in hbmenu. That runs as an applet, right?

leo60228: It works with an application, but crashes the system with code 2144-0001 in hbmenu. I don't know how to test anything else.

yellows8: Add a comment regarding that this can be only used with AppletType_Application.

yellows8:
Change to: appletSelfExit

nx/include/switch/services/applet.h
leo60228: Okay. Should I squash?

yellows8: doesn't matter

leo60228:
Done.

yellows8:
TBH it would be better to have libnx handle calling this internally instead of the user-app, but not sure how to handle that.

yellows8:
beef09a27e Actually had the code for this in applet.c already.

PR #175, opened 2018-09-21
Reduce usage of malloc in select and poll. Fix poll not acceping -1 fds

TuxSH:
Not tested (can't test rn)

### DIFF
--- a/nx/source/runtime/devices/socket.c
+++ b/nx/source/runtime/devices/socket.c
@@ -4,6 +4,7 @@
 #include <ctype.h>
 #include <string.h>
 #include <malloc.h>
+#include <alloca.h>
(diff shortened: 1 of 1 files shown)


fincs:
Why not use alloca instead of tls?

TuxSH:
Because user has to know that he has to have a big enough stack.

Putting it in TBSS enforces reserving enough memory by libnx thread functions.

fincs:
The stack on the Switch is bigger though, and POLLFD_SB_MAX_FDS could be tweaked as to not consume that much stack. Putting it in tbss means *every single thread created* gets a buffer allocated for this, regardless of whether it uses select/poll or not.

PR #178, opened 2018-09-27
hid: Call SetSupportedNpadStyleSet to enable additional Npad features

endrift:
This enables additional Npad features, so StartSixAxisSensor will actually do things other than just log timestamps.

### DIFF
--- a/nx/include/switch/services/hid.h
+++ b/nx/include/switch/services/hid.h
@@ -592,6 +592,9 @@ void hidJoystickRead(JoystickPosition *pos, HidControllerID id, HidControllerJoy
 /// Returns 0 when CONTROLLER_PLAYER_1 is connected, otherwise returns 1 for handheld-mode.
 bool hidGetHandheldMode(void);
 
+/// Sets which controller types are supported. This is automatically called with all types in \ref hidInitialize.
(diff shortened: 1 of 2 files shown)


yellows8:
Can you call that from a dedicated func + expose it via hid.h?(Use HidControllerType for input and mention the automatic usage in comments)

endrift:
The input is a NpadStyleTag, not a HidControllerType, and I don't have any idea what the values for that are. This is just what BotW calls this function with.

yellows8:
NpadStyleTag == HidControllerType, latter is unofficial name.

endrift:
Oh ok, I'll do that.

nx/include/switch/services/hid.h
```diff
@@ -592,6 +592,9 @@ void hidJoystickRead(JoystickPosition *pos, HidControllerID id, HidControllerJoy
 /// Returns 0 when CONTROLLER_PLAYER_1 is connected, otherwise returns 1 for handheld-mode.
 bool hidGetHandheldMode(void);
 
+/// Sets which controller types are supported. This is automatically called with all types in hidInit.
```
yellows8: hidInitialize

PR #183, opened 2018-10-09
add an exit request hook type, allow to delay exit

p-sam:
This will permit to be able to be notified on exit by:
- hooking on exit
- being able to delay exit until decided

Here is an example of use: https://gist.github.com/p-sam/bd39c9a084cfd45c793d0bdfe2b8f2c7

Homebrew would be able to use this moment to flush state, and close handles. (eg: RetroArch flushing SaveRAM)

### DIFF
--- a/nx/source/services/applet.c
+++ b/nx/source/services/applet.c
@@ -1211,6 +1211,7 @@ bool appletMainLoop(void) {
     switch(msg) {
         case 0x4:
             appletCallHook(AppletHookType_OnExitRequest);
+            return false;
         break;
 
         case 0xF:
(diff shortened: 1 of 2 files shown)


nx/source/services/applet.c
```diff
@@ -1211,6 +1209,10 @@ bool appletMainLoop(void) {
     }
 
     switch(msg) {
+        case 0x4:
+            appletCallHook(AppletHookType_OnExitRequest);
+        break;
```
yellows8: Add: <code>return false;</code> here. Apps won't need to use appletHook with this, just use appletUnlockExit after main-loop exit. Apps should *not* exit without calling appletUnlockExit if appletLockExit was used.

p-sam: true, that makes that use-case much simpler, I also updated the gist to reflect change

yellows8:
Can you doc the lock/unlock cmds on switchbrew? https://switchbrew.org/wiki/AM_services#LockExit

p-sam:
> Can you doc the lock/unlock cmds on switchbrew? https://switchbrew.org/wiki/AM_services#LockExit

sure, will do i a bit

PR #185, opened 2018-10-15
Make FS use domains.

SciresM:
Adds new serviceIpcPrepareHeader, serviceIpcParse, serviceCreateSubservice functions.

Causes a breaking (but necessary) change to ipcParseForDomain.

Makes fsInitialize() call serviceConvertToDomain(), resulting in all FS commands using domain object IDs.

### DIFF
--- a/nx/include/switch/services/sm.h
+++ b/nx/include/switch/services/sm.h
@@ -16,7 +16,7 @@ typedef enum {
     ServiceType_Normal,             ///< Normal service.
     ServiceType_Domain,             ///< Domain.
     ServiceType_DomainSubservice,   ///< Domain subservice;
-    ServiceType_Override            ///< Service overriden in the homebrew environment.
+    ServiceType_Override,           ///< Service overriden in the homebrew environment.
 } ServiceType;
 
 /// Service object structure.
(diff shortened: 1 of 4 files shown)


SciresM:
Ideally over time we can refactor other services to use these primitives. 

QuestionL Should we eventually break ipcParse() and make it take a size? It would be nice to have the RawSize be more meaningful in IpcParsedCommand...

SciresM:
Oops, just realized this actually breaks IpcParsedCommand support for domain on the serverside. Should be straightforward, will push a fix tomorrow morning.

nx/include/switch/services/sm.h
```diff
@@ -113,6 +113,20 @@ static inline void serviceCreateDomainSubservice(Service* s, Service* parent, u3
     s->object_id = object_id;
 }
 
+/**
+ * @brief Creates a domain subservice object from a parent service.
```
yellows8: Update these comments, it supports regular serviceCreate too.

nx/include/switch/services/sm.h
```diff
@@ -113,6 +113,20 @@ static inline void serviceCreateDomainSubservice(Service* s, Service* parent, u3
...
+ * @param[out] s Service object.
+ * @param[in] parent Parent service, necessarily a domain or domain subservice.
+ * @param[in] object_id Object ID for this subservice.
```
yellows8: This param doesn't exist & the r/i params are missing in these comments.

yellows8: This param doesn't exist & the r/i params are missing in these comments.

PR #190, opened 2018-10-24
set:sys: Add GetFirmwareVersion(2)

SciresM:


nx/include/switch/services/set.h
```diff
@@ -64,6 +64,22 @@ typedef enum {
     SetSysFlag_HeadphoneVolumeUpdate = 117,
 } SetSysFlag;
 
+/// Structure returned by GetFirmwareVersion(2)
```
yellows8: *by \ref setsysGetFirmwareVersion.

PR #192, opened 2018-10-26
fix print \n twice

SegFault42:
if '\n' is printed at windowWidth + 1, consoleNewRow() will be executed twice.

Example:

```
printf("%s\n", "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa");

```

80 'a' will be printed, then new line because the cursor is a the end of the line, and then our '\n' will be printed

### DIFF
--- a/nx/source/runtime/devices/console.c
+++ b/nx/source/runtime/devices/console.c
@@ -625,7 +625,8 @@ void consolePrintChar(int c) {
 	if(currentConsole->cursorX  >= currentConsole->windowWidth) {
 		currentConsole->cursorX  = 0;
 
-		consoleNewRow();
+		if (c != '\n')
+			consoleNewRow();
 	}
 
 	switch(c) {


chaoskagami:
I'm relatively sure that's not a bug and is in fact the intended behavior of ANSI escape sequences when being output to a vt. You can reproduce this "bug" on any modern OS with a terminal available.

SegFault42:
tested on my term
```
#include <stdio.h>

int	main()
{
	printf("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa\nbbb\n");
}

```
```
#include <stdio.h>

int	main()
{
	printf("aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaabbb\n");
}

```



Same output with '\n' and without '\n' at end of line

SegFault42:
found this function (early_vga_write()): 
https://elixir.bootlin.com/linux/latest/source/arch/x86/kernel/early_printk.c

PR #193, opened 2018-10-27
Timers and intra-process events

plutooo:


fincs:
(GitHub doesn't seem to let me comment on a commit on a PR)
fa9f933 <- I think this will break the self-relocation runtime code (which runs before anything else is initialized) - it uses fatalSimple and fatalSimple needs to test the kernel version to use a suitable fatal-error mode. Also, since it's adding a reference to an external (non-static) variable, accesses will go through GOT - which isn't fixed yet...

nx/include/switch/arm/counter.h
```diff
@@ -26,3 +26,11 @@ static inline u64 armGetSystemTickFreq(void) {
     __asm__ ("mrs %x[data], cntfrq_el0" : [data] "=r" (ret));
     return ret;
 }
+
+static inline u64 armNsToTick(u64 ns) {
```
fincs: armNsToTick/armTickToNs and the changes to event.c should be a separate commit, and this commit should be pushed directly to libnx outside this PR.

nx/source/kernel/utimer.c
```diff
@@ -0,0 +1,35 @@
...
+    t->next_time = 0;
+    t->interval = armNsToTick(interval);
+
+    if (start)
+        utimerStart(t);
+}
+
+void utimerStart(UsermodeTimer* t)
+{
+    __sync_bool_compare_and_swap(&t->next_time, 0, armGetSystemTick() + armNsToTick(t->interval));
```
fincs: __sync_* functions are deprecated GCC intrinsics. Please replace them with the modern [__atomic_*](https://gcc.gnu.org/onlinedocs/gcc/_005f_005fatomic-Builtins.html) functions. Also, compare-and-swap is an operation that should be in a loop, or else this is broken.

nx/include/switch/kernel/uevent.h
```diff
@@ -0,0 +1,23 @@
...
+    WaiterNode waiter_list;
+    bool signal;
+    bool auto_clear;
+};
+
+void ueventCreate(UsermodeEvent* e, bool auto_clear);
+void ueventClear(UsermodeEvent* e);
+void ueventSignal(UsermodeEvent* e);
+
+// Internal methods (do not use!):
```
fincs: Internal methods should not be in public header files. These should be in an internal header inside source/kernel/.

nx/include/switch/kernel/utimer.h
```diff
@@ -0,0 +1,19 @@
...
+{
+    u64 next_time;
+    u64 interval;
+};
+
+void utimerCreate(UsermodeTimer* t, u64 interval, bool start);
+void utimerStart(UsermodeTimer* t);
+void utimerStop(UsermodeTimer* t);
+
+// Internal methods (do not use!):
```
fincs: Same here

nx/include/switch/kernel/waiter.h
```diff
@@ -0,0 +1,35 @@
...
+{
+    Mutex mutex;
+    Handle thread;
+    s32 signalled_idx;
+
+    WaiterNode nodes[MAX_WAIT];
+    size_t num_nodes;
+};
+
+// Internal methods (do not use!):
```
fincs: Same here

nx/include/switch/kernel/uevent.h
plutooo: We don't use this pattern elsewhere

fincs: We do, it's internal.h.

PR #198, opened 2018-10-30
Allow passing descriptor to usbCommsInitializeEx

ConnorRigby:


SciresM:
@ConnorRigby Why? This isn't really semantically correct because of the way string descriptors work on >= 5.0.0. Doesn't also really make sense for the use-case for usbComms, which should have a static device descriptor anyway since it's always the same capability.

yellows8:
Remember that depending on a different device_descriptor == pre-5.0.0 can't be supported.

yellows8:
Can you ignore device_descriptor and detect it with the interface instead?

misson20000:
You can change certain fields between 2.0.0 and 5.0.0, but on 1.0.0 you're locked to `NintendoSdkDebugger` VID/PID of `0x057e`/`0x3000` (which I might note is the same VID/PID that @ConnorRigby wants and is perhaps why he wants to change the VID/PID in the first place so that his application is consistent on all firmware versions)

SciresM:
@misson20000 Yeah, we should already be consistent across firmware versions...

yellows8:
With custom host-side apps there's no point in this (?), VID/PID is constant already. If there's an actual use case for changing VID/PID perhaps it could be impl'd without overriding the entire descriptor.

PR #200, opened 2018-11-03
psc: Add PmModule support

SciresM:


nx/source/services/psc.c
```diff
@@ -0,0 +1,273 @@
...
+            if (R_FAILED(rc)) {
+                serviceClose(&out->srv);
+            }
+        }
+    }
+
+    return rc;
+}
+
+
```
yellows8: extra newline

nx/include/switch/services/psc.h
```diff
@@ -0,0 +1,34 @@
...
+typedef struct {
+    Event event;
+    Service srv;
+    u16 module_id;
+} PscPmModule;
+
+Result pscInitialize(void);
+void pscExit(void);
+
+Result pscGetPmModule(PscPmModule *out, u16 module_id, const u16 *dependencies, size_t dependency_count, bool autoclear);
```
fincs: Is autoclear really part of the abi? Are there PmModules which would want autoclear and others which would not?

PR #201, opened 2018-11-07
Expose appletSetFocusHandlingMode

natinusala:
This allows homebrew to enable or disable suspending when losing focus.


yellows8:
What's the use case for disabling suspending?

natinusala:
RetroArch doesn't like being suspended.

Instead with this I can use the AppletHook to detect loss of focus and pause the core instead (or not, user's choice) but keep the homebrew running in the background. 

I can then resume the core when I regain focus.

yellows8:
Be more specific with "RetroArch doesn't like being suspended."

natinusala:
- Potentially long background tasks are stopped : content scanner, database updater, thumbnails updater, assets updater...
- Threaded audio doesn't resume properly and is bad once resumed ("rice crispies")
- Not suspending allows using the HOME menu while being in a netplay session without terminating it, or just while waiting for a long download to end

yellows8:
"are stopped" You want this to continue running while out-of-focus?

Makes sense for networking.

yellows8:
However, properly fixing audio would be ideal...

yellows8:
Anyway, there should be enums for appletSetFocusHandlingMode / appletGetFocusState - apps shouldn't be using hard-coded constants for those (if it's known what the values for those are).

natinusala:
`appletGetFocusState`: 1 is focused, 3 is out of focus (home, sleep), 2 is unknown but it's probably because of an overlay applet (software keyboard, browser, controllers selection menu). Do you want me to leave 2 as unknown or do you want me to assume it's for overlays ?

`appletSetFocusHandlingMode`: the mode can either be 0, 1, 2 or 3. It changes values for the three booleans passed in the function, but it's not documented anywhere, do you know what these values are for (`invals` array) ?

yellows8:
"leave 2 as unknown or do you want me to assume it's for overlays ?" former

"do you know what these values are " nope

natinusala:
```
typedef enum {
    AppletFocusState_Focused = 1,               ///< Applet is focused.
    AppletFocusState_NotFocusedUnknown = 2,     ///< Out of focus - unkown reason (browser / keyboard /controllers menu ?).
    AppletFocusState_NotFocusedHomeSleep = 3    ///< Out of focus - HOME menu is open / console is sleeping.
} AppletFocusState;
 typedef enum {
  AppletFocusHandlingMode_Suspend1 = 0, ///< Suspend when out of focus.
  AppletFocusHandlingMode_NoSuspend,    ///< Don't suspend when out of focus.
  AppletFocusHandlingMode_Suspend2,     ///< Suspend when out of focus (unknown variant 1).
  AppletFocusHandlingMode_Suspend3      ///< Suspend when out of focus (unknown variant 2).
} AppletFocusHandlingMode;
```
This doesn't look pretty good but that's the best I can do so far while we find out what all the modes do.

nx/include/switch/services/applet.h
```diff
@@ -32,6 +32,21 @@ typedef enum {
     AppletHookType_Max,               ///< Number of applet hook types.
 } AppletHookType;
 
+typedef enum {
+    AppletFocusState_Focused = 1,               ///< Applet is focused.
+    AppletFocusState_NotFocusedUnknown = 2,     ///< Out of focus - unkown reason (browser / keyboard /controllers menu ?).
```
yellows8: *unknown reason (LibraryApplet open?)

nx/source/services/applet.c
```diff
@@ -1365,8 +1364,8 @@ u32 appletGetPerformanceMode(void) {
     return g_appletPerformanceMode;
 }
 
-u8 appletGetFocusState(void) {
-    return g_appletFocusState;
+AppletFocusState appletGetFocusState(void) {
+    return (AppletFocusState) g_appletFocusState;
```
yellows8: (AppletFocusState)g_appletFocusState;

PR #202, opened 2018-11-09
fs: Add fsFsQueryEntry (and fsFsSetArchiveBit helper)

SciresM:
This adds fsFsQueryEntry, an IFileSystem command added in 4.0.0.

It also adds a helper for using fsFsQueryEntry to set the archive bit. Please note, HOS returns 0x202 when calling SetArchiveBit on a file (and only allows setting on directories). This should be fine, though, because HOS only assigns special meaning to the archive bit for directories.


yellows8:
"This should be fine, though, because HOS only assigns special meaning to the archive bit for directories." Files with that bit set aren't listed in dir-listings though?

There should be fsdev wrapper func(s) for fsFsSetArchiveBit, not sure how it should be handled.

yellows8:
This also needs a sysver check.

PR #220, opened 2018-12-09
Implement various services for debug monitor usage.

SciresM:
This PR adds ro:dmnt, and fully fleshes out ns:dev, pm:dmnt, pm:shell. It also adds ldr:ro, since that was apparently missing..

### DIFF
--- a/nx/include/switch/services/ns.h
+++ b/nx/include/switch/services/ns.h
@@ -8,6 +8,7 @@
 #include "../types.h"
 #include "../nacp.h"
 #include "../services/fs.h"
+#include "../kernel/event.h"
 
 typedef struct {
     NacpStruct nacp;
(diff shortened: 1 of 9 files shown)


nx/include/switch/services/ns.h
```diff
@@ -34,6 +35,28 @@ typedef struct
...
+typedef enum {
+    NsShellEvent_None = 0,
+    NsShellEvent_Exit = 1,
+    NsShellEvent_Start = 2,
+    NsShellEvent_Crash = 3,
+    NsShellEvent_Debug = 4,
+} NsShellEvent;
+
+typedef struct {
+    NsShellEvent event : 32;
```
fincs: The `: 32` isn't necessary, because on AArch64 enums are at least 32-bit.

nx/source/services/ro.c
```diff
@@ -0,0 +1,284 @@
...
+#include "arm/atomics.h"
+#include "kernel/ipc.h"
+#include "kernel/detect.h"
+#include "services/ro.h"
+#include "services/sm.h"
+
+static Service g_roSrv, g_dmntSrv;
+static u64 g_roRefCnt, g_dmntRefCnt;
+
+Result _ldrRoInitialize(void);
```
fincs: Should be declared as static

nx/source/services/ro.c
```diff
@@ -0,0 +1,284 @@
...
+        serviceIpcParse(&g_roSrv, &r, sizeof(*resp));
+        resp = r.Raw;
+
+        rc = resp->result;
+    }
+
+    return rc;
+}
+
+Result _ldrRoInitialize(void) {
```
yellows8: static

nx/include/switch/services/ns.h
```diff
@@ -65,5 +87,15 @@ Result nsvmGetSafeSystemVersion(u16 *out);
 Result nsdevInitialize(void);
 void nsdevExit(void);
 
+Result nsdevLaunchProgram(u64* out_pid, const NsLaunchProperties* properties, u32 flags);
 Result nsdevTerminateProcess(u64 pid);
 Result nsdevTerminateProgram(u64 tid);
+Result nsdevGetShellEvent(Event* out);
```
yellows8: Add comment for autoclear.

nx/include/switch/services/pm.h
```diff
@@ -18,14 +42,23 @@ void pminfoExit(void);
...
 Result pmdmntGetApplicationPid(u64* pid_out);
 Result pmdmntEnableDebugForApplication(Handle* handle_out);
+Result pmdmntDisableDebug(void);
 
 Result pminfoGetTitleId(u64* title_id_out, u64 pid);
 
 Result pmshellLaunchProcess(u32 launch_flags, u64 titleID, u64 storageID, u64 *pid);
+Result pmshellTerminateProcessByProcessId(u64 processID);
 Result pmshellTerminateProcessByTitleId(u64 titleID);
+Result pmshellGetProcessEvent(Event* out);
```
yellows8: Add comment for autoclear.

nx/source/services/ro.c
fincs: static only needs to be placed in the prototype, not in the function definition. In fact you get a warning iirc.

PR #223, opened 2018-12-17
Add missing applet focus states and handling modes

natinusala:
I added the missing applet focus states and handling modes in `applet.h`.

- The unknown focus state is indeed because of an open LibraryApplet
- Suspend1 is now SuspendHomeSleep, as it only suspends when the HOME menu is open or the console is sleeping
- Suspend2 is now SuspendHomeSleepNotify, as it only suspends when the HOME menu is open or the console is sleeping while still recieving the OnFocusState hook
- Suspend3 is now AlwaysSuspend, as it always suspends the applet regardless of the reason (HOME menu or LibraryApplet)

### DIFF
--- a/nx/include/switch/services/applet.h
+++ b/nx/include/switch/services/applet.h
@@ -35,18 +35,18 @@ typedef enum {
 } AppletHookType;
 
 typedef enum {
-    AppletFocusState_Focused = 1,               ///< Applet is focused.
-    AppletFocusState_NotFocusedUnknown = 2,     ///< Out of focus - unknown reason (LibraryApplet open?).
-    AppletFocusState_NotFocusedHomeSleep = 3    ///< Out of focus - HOME menu is open / console is sleeping.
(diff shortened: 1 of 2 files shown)


CTCaer:
It seems that they follow a pattern:

  | SuspendHomeSleep | AlwaysSuspend | NoSuspend | SuspendHomeSleepNotify
-- | -- | -- | -- | --
invals[0] | FALSE | FALSE | TRUE | TRUE
invals[1] | FALSE | FALSE | TRUE | FALSE
invals[2] | TRUE | TRUE | FALSE | TRUE
invals[3] | FALSE | TRUE | FALSE | FALSE
  |   |   |   |  
invals[0] | OnFocusStateNotify |   |  
invals[1] | NoSuspendOnSleep? |   |  
invals[2] | HomeSuspendEnable |   |  
invals[3] | OutOfFocusSuspend |   |  

Based on these we can also use a struct for the invals.

PR #227, opened 2018-12-29
Mutex Error Checking

RSDuck:
I'm currently trying to implement pthreads as good as possible (or rather as good as I'm eable to pull off). To implement all [mutex types](https://pubs.opengroup.org/onlinepubs/7908799/xsh/pthread_mutex_lock.html), I need to know whether a mutex is already owned by the current thread when locking it, aswell as if it even belongs to the current thread, when unlocking.

### DIFF
--- a/nx/include/switch/kernel/mutex.h
+++ b/nx/include/switch/kernel/mutex.h
@@ -26,8 +26,9 @@ static inline void mutexInit(Mutex* m)
 /**
  * @brief Locks a mutex.
  * @param m Mutex object.
+ * @return 0 if the mutex is already owned by this thread, and 1 if it was sucessfully acquired.
  */
-void mutexLock(Mutex* m);
+bool mutexLock(Mutex* m);
 
 /**
  * @brief Attempts to lock a mutex without waiting.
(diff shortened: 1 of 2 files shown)


fincs:
> I'm currently trying to implement pthreads as good as possible

This is something that needs proper support from the toolchain (devkitA64), and thus any attempts to implement pthreads that do not involve making the necessary support changes to newlib will unfortunately be doomed to either fail or produce an unmaintainable solution. In addition, the prototypes of mutexUnlock/rmutexUnlock must match the prototypes in the __syscalls structure (you can see how this PR fails to compile in the doozer build log). Therefore this PR cannot be accepted.

fincs:
Before we get to pthreads, we'd like to implement support for `std::thread` first; there's already an approach that we're planning to follow that is similar to what would be needed for pthread, and that involves GCC's own internal threading lib (gthread).

RSDuck:
I actually implemented a [subset](https://github.com/RSDuck/nxcppthread) of ``std::thread`` before realising that pthreads would the better choice. As far as I perceived it, gthreads are implemented on top of pthreads(as done in gthr-posix.h), but of course it also works the other way around.

fincs:
It's kind of frustrating when this stuff keeps on happening, and people partake in attempts to implement crucial parts of the toolchain without trying to talk to the maintainers of the toolchain first. I had a look at the lib you wrote, and I can tell it's going to break pretty bad once we have proper support for `std::thread` working in libstdc++/libnx.

RSDuck:
I fully understand what you're meaning, I should probably checkout #switchdev more often :)
This obviously is open source software, so it isn't hard planned and organised or anything like that, though I would be very greatful to see some kind of roadmap on the future of this toolchain

PR #229, opened 2019-01-02
Implement a way to set the NRO that the next NRO will return to

friedkeenan:
This is to be used with [my fork of nx-hbloader](https://github.com/friedkeenan/nx-hbloader) (Pull request [here](https://github.com/switchbrew/nx-hbloader/pull/12)).

I made this in about 3 and a half hours while tired, so please cut me some slack if I did anything absolutely terribly.

You can see a simple example [here](https://gist.github.com/friedkeenan/fdef33762dbd3f5e09cea8b6d7ad38f0).

The variable and function names aren't very good in my opinion, but I didn't know what else to call them.

I mainly did this because I thought it was not ideal that [Goldleaf](https://github.com/XorTroll/Goldleaf) had to return to the homebrew menu after launching an NRO, and couldn't return back to itself.

### DIFF
--- a/nx/include/switch/runtime/env.h
+++ b/nx/include/switch/runtime/env.h
@@ -34,6 +34,7 @@ enum {
     EntryType_ProcessHandle=10,       ///< Provides the process handle.
     EntryType_LastLoadResult=11,      ///< Provides the last load result.
     EntryType_RandomSeed=14,          ///< Provides random data used to seed the pseudo-random number generator.
+    EntryType_NroReturnPath=15
 };
 
 enum {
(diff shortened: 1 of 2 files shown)


friedkeenan:
So, it's been a while, and I was wondering if you guys have anything to say about this, or have talked about it internally? I'm sorry if I seem impatient or anything like that, but I'm just curious what your thoughts are about this

friedkeenan:
Gonna close this since nothing has really happened with it; hopefully I'll code up a better solution (probably using a stack thing) and make a new PR

PR #232, opened 2019-01-09
c11-threads: return thrd_timedout on cnd_timedwait timeout

Cpasjuste:


nx/source/runtime/c11-threads.c
```diff
@@ -114,6 +114,10 @@ static int __cnd_timedwait(cnd_t *__restrict cond, mtx_t *__restrict mtx, u64 ti
         mtx->rmutex.counter = 1;
     }
 
+    if (rc == 0xEA01) {
```
fincs:
```c
if (R_VALUE(rc) == KERNELRESULT(TimedOut))
    return thrd_timeout;
```

PR #234, opened 2019-01-24
Added initial implementation of es services

simontime:
This initial implementation provides access to both `esCountCommonTicket` and `esCountPersonalizedTicket`.


XorTroll:
Personally, why would people want to mess up with tickets, like it's understandable for some homebrew, but...

simontime:
Hence why this provides access to the two safest `Get`-related services only, to ensure people don't accidentally screw up their ticket db.

simontime:
Is there a specific reason why this was closed?

fincs:
There's no meaningful use case for ticket functions, other than supporting piracy-related tools.

PR #255, opened 2019-03-23
rwlock: Move over to condvar implementation.

plutooo:
This one is superior to the previous because it is write-preferring.

### DIFF
--- a/nx/include/switch/kernel/rwlock.h
+++ b/nx/include/switch/kernel/rwlock.h
@@ -13,8 +13,8 @@ typedef struct {
     Mutex mutex;
     CondVar condvar_readers;
     CondVar condvar_writer;
-    u64  readers;
-    bool writer;
+    u32  readers : 31;
+    bool writer : 1;
 } RwLock;
 
 /**
--- a/nx/include/switch/kernel/rwlock.h
+++ b/nx/include/switch/kernel/rwlock.h
@@ -6,12 +6,15 @@
  */
 #pragma once
 #include "../kernel/mutex.h"
+#include "../kernel/condvar.h"
 
 /// Read/write lock structure.
 typedef struct {
-    RMutex r;
-    RMutex g;
-    u64 b;
+    Mutex mutex;
+    CondVar condvar_readers;
+    CondVar condvar_writer;
+    u64  readers;
+    bool writer;
 } RwLock;
 
 /**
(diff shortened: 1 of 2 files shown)


nx/include/switch/kernel/rwlock.h
```diff
@@ -6,12 +6,15 @@
...
 /// Read/write lock structure.
 typedef struct {
-    RMutex r;
-    RMutex g;
-    u64 b;
+    Mutex mutex;
+    CondVar condvar_readers;
+    CondVar condvar_writer;
+    u64  readers;
+    bool writer;
```
fincs: Save space and avoid alignment padding by using bitfields:
```c
    u32 readers : 31;
    bool writer : 1;
```

nx/source/kernel/rwlock.c
```diff
@@ -3,33 +3,57 @@
...
+    mutexUnlock(&r->mutex);
 }
 
 void rwlockReadUnlock(RwLock* r) {
-    rmutexLock(&r->r);
+    mutexLock(&r->mutex);
 
-    if (r->b-- == 1)
-        rmutexUnlock(&r->g);
+    if (r->readers-- == 1) {
```
fincs: `if (!--r->readers)`

nx/source/kernel/rwlock.c
```diff
@@ -3,33 +3,57 @@
...
-    rmutexLock(&r->g);
+    mutexLock(&r->mutex);
+
+    while (r->writer) {
+        condvarWait(&r->condvar_writer, &r->mutex);
+    }
+
+    r->writer = true;
+
+    while (r->readers > 0) {
```
fincs: `while (r->readers)`

fincs:
Pinging @plutooo.

nx/include/switch/kernel/rwlock.h
plutooo: fixed

nx/source/kernel/rwlock.c
plutooo: fixed-ish, prefer not using boolean operators (!) on integer types

nx/source/kernel/rwlock.c
plutooo: prefer not using boolean operators (!) on integer types

nx/source/kernel/rwlock.c
fincs: That's fine. However, those parentheses are superflous and you can remove them: `if (--r->readers == 0)` is also correct.

PR #260, opened 2019-04-06
Avoid inline functions in IPC

XorTroll:
Rust's bindgen cannot generate properly inline functions' bindings, so this would be really helpful for Rust <-> libnx bindings.

### DIFF
--- a/nx/include/switch/kernel/ipc.h
+++ b/nx/include/switch/kernel/ipc.h
@@ -97,9 +97,7 @@ typedef struct {
  * @brief Initializes an IPC command structure.
  * @param cmd IPC command structure.
  */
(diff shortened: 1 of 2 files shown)


fincs:
We don't support Rust or any other languages positioning themselves to replace C/C++. People attempting to use non-C/C++ languages cause endless support problems for us and we would rather have people stick to standard tools and libraries. libnx IPC system heavily relies on inline functions for good code generation and optimization, and many other things would be severely impacted by this PR. Closing.

roblabla:
FWIW: The rust binding could reexport the static symbols by making a shim C library that exports the symbols (e.g. create a bunch of functions that just call the inlined version). The rust side will obviously have not-great perfs this way, but if performance is not critical, it's an OK solution.

PR #262, opened 2019-04-15
kernel/thread: Silence -Wsign-compare warnings

lioncash:
Ensures that the same signedness of type is being used in comparisons where applicable

### DIFF
--- a/nx/source/kernel/thread.c
+++ b/nx/source/kernel/thread.c
@@ -138,7 +138,7 @@ void threadExit(void) {
         fatalSimple(MAKERESULT(Module_Libnx, LibnxError_NotInitialized));
 
     u64 tls_mask = __atomic_load_n(&g_tlsUsageMask, __ATOMIC_SEQ_CST);
-    for (s32 i = 0; i < NUM_TLS_SLOTS; i ++) {
+    for (size_t i = 0; i < NUM_TLS_SLOTS; i ++) {
         if (!(tls_mask & ((UINT64_C(1) << i))))
             continue;
         if (t->tls_array[i]) {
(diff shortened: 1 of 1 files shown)


fincs:
I'm not getting warnings in thread.c. How did you trigger those warnings?

```
thread.c
aarch64-none-elf-gcc -MMD -MP -MF /home/fincs/Projects/Switch/libnx/nx/release/thread.d -g -Wall -Werror -ffunction-sections -fdata-sections -march=armv8-a+crc+crypto -mtune=cortex-a57 -mtp=soft -fPIC -ftls-model=local-exec -DNDEBUG=1 -O2 -I/home/fincs/Projects/Switch/libnx/nx/include -I/home/fincs/Projects/Switch/libnx/nx/external/bsd/include  -I. -iquote /home/fincs/Projects/Switch/libnx/nx/include/switch/ -DSWITCH -c /home/fincs/Projects/Switch/libnx/nx/source/kernel/thread.c -o thread.o
```

lioncash:
Ah, yeah, I made changes locally to the makefile when testing some stuff and accidentally left those changes in when revisiting things, my bad.

Tangentially, since it's related to warnings; in external C++ projects with elevated warnings levels (i.e. using -Wextra), [this](https://github.com/switchbrew/libnx/blob/master/nx/include/switch/kernel/ipc.h#L101) line seems to cause quite a lot of missing initializer warnings.

e.g.

warnings

is there a preferable way I could resolve this upstream? My initial thought was to just open a PR using memset, but given it's in a header, I assume including string.h is undesirable here (and specifying all the initializers may seem gross as well).

fincs:
As a general amount, we don't usually fix warnings stemming from using `-Wextra`, since that flag is well known to cause lots of spurious and/or false positive warnings. I suppose `__builtin_memset` could be used here. I'm not sure, but would just `(IpcCommand){}` suffice?

lioncash:
Yeah, `(IpcCommand){}` indeed works.

fincs:
Ok so go ahead. I'll be closing this one in the meantime.

PR #270, opened 2019-04-25
sm: add retry loop in smInitialize

SciresM:
This matches official software behavior, and is necessary for FIRM sysmodules.


nx/source/services/sm.c
```diff
@@ -56,8 +56,12 @@ Result smInitialize(void)
         return 0;
 
     Result rc = svcConnectToNamedPort(&g_smHandle, "sm:");
-    Handle tmp;
+    while (rc == 0xF201) {
```
fincs: `while (R_VALUE(rc) == KERNELRESULT(NotFound))`

PR #273, opened 2019-05-02
i2c: flesh out I2cDevice enum

SciresM:
Renames _AudioCodec for standardization reasons.


nx/include/switch/services/i2c.h
```diff
@@ -9,7 +9,39 @@
 #include "sm.h"
 
 typedef enum {
-    I2cDevice_AudioCodec = 4,
+    I2cDevice_DebugPad = 0,
```
yellows8: Align '=' on each line, like elsewhere.

PR #278, opened 2019-05-09
ncm: add ncmContentMetaDatabaseGetAttributes

SciresM:
Implements ncm::IContentMetaDatabase->Cmd18


SciresM:
Also just added enums for title type/result attribute.

PR #281, opened 2019-05-17
i2c: add ReceiveAudo and ExecuteCommandList

Thog:
Also make buf const in SendAuto.


yellows8:
Doc i2csessionReceiveAuto on wiki?

PR #282, opened 2019-05-30
Add R_TRY helper macro.

SciresM:
This adds a macro that evaluates a result expression, and returns it on failure.

Greatly reduces if (R_FAILED((rc = )) boilerplate.


SciresM:
Closing per discussion (undesirable to add complex macros).

PR #288, opened 2019-06-02
Implement fsOpenContentStorageFileSystem, fsGetRightsIdByPath & fsGetRightsIdAndKeyGenerationByPath

Adubbz:
Required for NCM reimplementation. Please note that the RightsId struct had to be moved from the NCM header to the FS header as they both rely on it, and the NCM header already depends on the FS header. Leaving it as-is would lead to a circular dependency, plus I believe it's more suitable for the FS header anyway.

### DIFF
--- a/nx/include/switch/services/fs.h
+++ b/nx/include/switch/services/fs.h
@@ -188,7 +188,11 @@ Result fsOpenDataStorageByCurrentProcess(FsStorage* out);
 Result fsOpenDataStorageByDataId(FsStorage* out, u64 dataId, FsStorageId storageId);
 Result fsOpenDeviceOperator(FsDeviceOperator* out);
 Result fsOpenSdCardDetectionEventNotifier(FsEventNotifier* out);
+
+// Retrieves the rights id corresponding to the content path. Only available on [2.0.0+].
 Result fsGetRightsIdByPath(const char* path, FsRightsId* out_rights_id);
+
+// Retrieves the rights id and key generation corresponding to the content path. Only available on [3.0.0+].
 Result fsGetRightsIdAndKeyGenerationByPath(const char* path, u8* out_key_generation, FsRightsId* out_rights_id);
 // todo: Rest of commands here
 
--- a/nx/include/switch/services/fs.h
+++ b/nx/include/switch/services/fs.h
@@ -21,6 +21,10 @@
 #define FS_SAVEDATA_USERID_COMMONSAVE 0
 
 typedef struct {
+    u8 c[0x10];
+} FsRightsId;
+
+typedef struct {
     Service  s;
 } FsFileSystem;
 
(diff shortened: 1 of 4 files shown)


nx/source/services/fs.c
```diff
@@ -465,6 +503,87 @@ Result fsOpenSdCardDetectionEventNotifier(FsEventNotifier* out) {
...
+            FsRightsId rights_id;
+        } *resp;
+
+        serviceIpcParse(&g_fsSrv, &r, sizeof(*resp));
+        resp = r.Raw;
+
+        rc = resp->result;
+
+        if (R_SUCCEEDED(rc)) {
+            *out_rights_id = resp->rights_id;
```
yellows8: `if ({out}) *{out} = ...`, same below.

nx/include/switch/services/fs.h
```diff
@@ -179,10 +183,13 @@ Result fsMountSdcard(FsFileSystem* out);
 Result fsMountSaveData(FsFileSystem* out, u8 inval, FsSave *save);
 Result fsMountSystemSaveData(FsFileSystem* out, u8 inval, FsSave *save);
 Result fsOpenSaveDataIterator(FsSaveDataIterator* out, s32 SaveDataSpaceId);
+Result fsOpenContentStorageFileSystem(FsFileSystem* out, FsContentStorageId content_storage_id);
 Result fsOpenDataStorageByCurrentProcess(FsStorage* out);
 Result fsOpenDataStorageByDataId(FsStorage* out, u64 dataId, FsStorageId storageId);
 Result fsOpenDeviceOperator(FsDeviceOperator* out);
 Result fsOpenSdCardDetectionEventNotifier(FsEventNotifier* out);
+Result fsGetRightsIdByPath(const char* path, FsRightsId* out_rights_id);
```
yellows8: Add comments for required sysvers, and check sysver.

nx/include/switch/services/fs.h
```diff
@@ -179,10 +183,17 @@ Result fsMountSdcard(FsFileSystem* out);
 Result fsMountSaveData(FsFileSystem* out, u8 inval, FsSave *save);
 Result fsMountSystemSaveData(FsFileSystem* out, u8 inval, FsSave *save);
 Result fsOpenSaveDataIterator(FsSaveDataIterator* out, s32 SaveDataSpaceId);
+Result fsOpenContentStorageFileSystem(FsFileSystem* out, FsContentStorageId content_storage_id);
 Result fsOpenDataStorageByCurrentProcess(FsStorage* out);
 Result fsOpenDataStorageByDataId(FsStorage* out, u64 dataId, FsStorageId storageId);
 Result fsOpenDeviceOperator(FsDeviceOperator* out);
 Result fsOpenSdCardDetectionEventNotifier(FsEventNotifier* out);
+
+// Retrieves the rights id corresponding to the content path. Only available on [2.0.0+].
```
yellows8: `///`, same below.

PR #289, opened 2019-06-03
Implement setsys GetDeviceNickname/SetDeviceNickname

thedax:
Implements nn::settings::ISystemSettingsServer commands 77 & 78.

https://reswitched.github.io/SwIPC/ifaces.html#nn::settings::ISystemSettingsServer(77)
https://reswitched.github.io/SwIPC/ifaces.html#nn::settings::ISystemSettingsServer(78)

Useful for apps that might want to display or set the console's nickname.

### DIFF
--- a/nx/include/switch/services/set.h
+++ b/nx/include/switch/services/set.h
@@ -184,6 +184,6 @@ Result setsysGetDeviceNickname(char* buffer);
 
 /**
  * @brief Sets the system's nickname.
- * @param buffer Pointer to read the nickname from. (The buffer size needs to be at least 0x80 bytes)
+ * @param buffer Pointer to read the nickname from.
  */
 Result setsysSetDeviceNickname(const char* buffer);
--- a/nx/include/switch/services/set.h
+++ b/nx/include/switch/services/set.h
@@ -186,4 +186,4 @@ Result setsysGetDeviceNickname(char* buffer);
  * @brief Sets the system's nickname.
  * @param buffer Pointer to read the nickname from. (The buffer size needs to be at least 0x80 bytes)
  */
-Result setsysSetDeviceNickname(char* buffer);
+Result setsysSetDeviceNickname(const char* buffer);
--- a/nx/include/switch/services/set.h
+++ b/nx/include/switch/services/set.h
@@ -11,6 +11,7 @@
 #include "../kernel/event.h"
 
 #define SET_MAX_NAME_SIZE 0x48
+#define SET_MAX_NICKNAME_SIZE 0x80
 
 typedef enum {
     ColorSetId_Light=0,
@@ -178,12 +179,12 @@ Result setsysGetFatalDirtyFlags(u64 *flags_0, u64 *flags_1);
 
 /**
  * @brief Gets the system's nickname.
- * @param buffer Pointer to output the nickname to. (The buffer size needs to be at least 0x80 bytes)
+ * @param nickname Pointer to output the nickname to. (The buffer size needs to be at least 0x80 bytes)
  */
-Result setsysGetDeviceNickname(char* buffer);
+Result setsysGetDeviceNickname(char* nickname);
 
 /**
  * @brief Sets the system's nickname.
- * @param buffer Pointer to read the nickname from.
+ * @param nickname Pointer to read the nickname from.
  */
-Result setsysSetDeviceNickname(const char* buffer);
+Result setsysSetDeviceNickname(const char* nickname);
--- a/nx/include/switch/services/set.h
+++ b/nx/include/switch/services/set.h
@@ -175,3 +175,15 @@ Result setsysBindFatalDirtyFlagEvent(Event *out);
  * @param flags_1 Pointer to populate with second 64 flags.
  */
 Result setsysGetFatalDirtyFlags(u64 *flags_0, u64 *flags_1);
+
+/**
+ * @brief Gets the system's nickname.
+ * @param buffer Pointer to output the nickname to. (The buffer size needs to be at least 0x80 bytes)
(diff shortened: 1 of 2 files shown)


nx/include/switch/services/set.h
```diff
@@ -175,3 +175,15 @@ Result setsysBindFatalDirtyFlagEvent(Event *out);
...
+ * @brief Gets the system's nickname.
+ * @param buffer Pointer to output the nickname to. (The buffer size needs to be at least 0x80 bytes)
+ */
+Result setsysGetDeviceNickname(char* buffer);
+
+/**
+ * @brief Sets the system's nickname.
+ * @param buffer Pointer to read the nickname from. (The buffer size needs to be at least 0x80 bytes)
+ */
+Result setsysSetDeviceNickname(char* buffer);
```
fincs: `const char*`

nx/include/switch/services/set.h
```diff
@@ -175,3 +175,15 @@ Result setsysBindFatalDirtyFlagEvent(Event *out);
...
+
+/**
+ * @brief Gets the system's nickname.
+ * @param buffer Pointer to output the nickname to. (The buffer size needs to be at least 0x80 bytes)
+ */
+Result setsysGetDeviceNickname(char* buffer);
+
+/**
+ * @brief Sets the system's nickname.
+ * @param buffer Pointer to read the nickname from. (The buffer size needs to be at least 0x80 bytes)
```
fincs: Does the buffer *really* need to be 0x80 bytes, or is this just copypasta error?

nx/source/services/set.c
```diff
@@ -686,3 +686,69 @@ Result setsysGetFatalDirtyFlags(u64 *flags_0, u64 *flags_1) {
...
+            u64 result;
+        } *resp = r.Raw;
+
+        rc = resp->result;
+    }
+
+    return rc;
+}
+
+Result setsysSetDeviceNickname(char* buffer) {
```
fincs: `const char*`

nx/source/services/set.c
```diff
@@ -686,3 +686,69 @@ Result setsysGetFatalDirtyFlags(u64 *flags_0, u64 *flags_1) {
...
+    IpcCommand c;
+    ipcInitialize(&c);
+
+    ipcAddSendBuffer(&c, buffer, 0x80, BufferType_Normal);
```
fincs: Hardcoded size for a string sounds fishy to me. Should this be `strlen(buffer)`? If it really needs to be 0x80 bytes in size, then a stack buffer with 0x80 bytes (initialized to zero through `= {0};`) should be used, and the string copied to it.

thedax: It doesn't seem to matter: I initially put 0x80 since SwIPC implied it needed that. However, when I passed it a 1 megabyte (1024 * 1024 bytes) buffer filled with 'A' and a null terminator at the end (after changing the above ipcAddSendBuffer to use strlen + 1), it didn't crash or return an error, so it seems to just truncate it.

yellows8:
0x80 is the official struct size.

nx/source/services/set.c
```diff
@@ -686,3 +686,69 @@ Result setsysGetFatalDirtyFlags(u64 *flags_0, u64 *flags_1) {
...
+    }
+
+    return rc;
+}
+
+Result setsysSetDeviceNickname(const char* buffer) {
+    IpcCommand c;
+    ipcInitialize(&c);
+
+    ipcAddSendBuffer(&c, buffer, strlen(buffer) + 1, BufferType_Normal);
```
yellows8: Can you copy `buffer` to a tmp 0x80-byte buf, then use that for ipcAddSendBuffer?

thedax: Like this? I followed the style from setsysGetSettingsItemValueSize. Should I also rename `buffer` to `nickname` and change `nickname` (the tmp buffer) to `send_nickname` for style consistency with how some of the other commands do it?

```c
// in set.h
#define SET_MAX_NICKNAME_SIZE 0x80

// in set.c
Result setsysSetDeviceNickname(const char* buffer) {    
    char nickname[SET_MAX_NICKNAME_SIZE];
    memset(nickname, 0, SET_MAX_NICKNAME_SIZE);
    strncpy(nickname, buffer, SET_MAX_NICKNAME_SIZE - 1);
    
    IpcCommand c;
    ipcInitialize(&c);
    ipcAddSendBuffer(&c, nickname, SET_MAX_NICKNAME_SIZE, 0);

    struct {
        u64 magic;
        u64 cmd_id;
    } *raw;

    raw = ipcPrepareHeader(&c, sizeof(*raw));

    raw->magic = SFCI_MAGIC;
    raw->cmd_id = 78;

    Result rc = serviceIpcDispatch(&g_setsysSrv);

    if (R_SUCCEEDED(rc)) {
        IpcParsedCommand r;
        ipcParse(&r);

        struct {
            u64 magic;
            u64 result;
        } *resp = r.Raw;

        rc = resp->result;
    }

    return rc;
}
```

yellows8: yeah

nx/source/services/set.c
```diff
@@ -720,11 +720,14 @@ Result setsysGetDeviceNickname(char* buffer) {
     return rc;
 }
 
-Result setsysSetDeviceNickname(const char* buffer) {
+Result setsysSetDeviceNickname(const char* nickname) {
+    char send_nickname[SET_MAX_NICKNAME_SIZE];
```
fincs: Instead of doing an explicit memset, do `char send_nickname[SET_MAX_NICKNAME_SIZE] = {0};`

PR #293, opened 2019-06-23
stub <sys/utsname.h>

switchjsdev:
POSIX Reference: https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/sys_utsname.h.html

Required entries:

- `struct utsname` with 5 required fields
- `int uname(struct utsname *);` which can be inline

### DIFF
--- a/nx/external/bsd/include/sys/utsname.h
+++ b/nx/external/bsd/include/sys/utsname.h
@@ -1,0 +1,31 @@
+/*
+ * Technically compliant implementation based on POSIX specification
+ */
+#ifndef _SYS_UTSNAME_H_
+#define _SYS_UTSNAME_H_
+
+/*
+ * "The character arrays are of unspecified size" assume 63 + null byte
+ *
+ * Members are stored in the order specified in the POSIX description
+ */
+struct utsname {
+  char sysname[64];
+  char nodename[64];
+  char release[64];
+  char version[64];
+  char machine[64];
+};
+
+/*
+ * `uname` can be defined as a macro or a function, so inline is fair game.
+ * 
+ * -1 shall be returned on error.
+ * 
+ * Note: errno should be set, but the spec doesn't give a list of valid codes.
+ */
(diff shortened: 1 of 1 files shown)


WinterMute:
This doesn't belong in the BSD headers and I'm not sure it's useful without a functioning uname. What's the goal of this PR?

switchjsdev:
Trying to port some code, the first noticeable missing header was `sys/utsname.h`.  The original library code defends against the `-1` return value, so this header-only implementation would be in compliance with POSIX without otherwise disrupting the rest of the codebase.

This was stuck in `bsd` because the dist build scripts already copy it and I didn't want to make a new path and disrupt the Makefile, also eventually the BSD version probably should be used.

fincs:
What code are you porting?

yellows8:
Pinging @switchjsdev

switchjsdev:
V8 javascript engine and octave math environment both require this header in the stack, in context both handle the -1 error case.  Hardcoding a result would also work.

@WinterMute FreeBSD provides its own https://github.com/freebsd/freebsd/blob/master/sys/sys/utsname.h

WinterMute:
v8 has no reference to utsname.h that I can find. Is there a repository for Octave I can look at?

switchjsdev:
v8 lives in the Chromium tree, and current builds from last known good revision require the "base" library, https://github.com/chromium/chromium/blob/master/base/system/sys_info_posix.cc for example.  On a side note, figuring out how to get the build system to play nice with the devkitpro toolchain was ... painful to say the least.

Octave source is hosted on savannah at https://hg.savannah.gnu.org/hgweb/octave/ , direct link to the include is https://hg.savannah.gnu.org/hgweb/octave/file/1ae91e5d70cc/liboctave/wrappers/uname-wrapper.c#l31

There are other headers and requirements to get them to compile.  Starting here since this is an innocuous header-only addition.

fincs:
Closing due to being out of scope for libnx in particular.

PR #302, opened 2019-07-04
Implement fsCreateSaveDataFileSystemBySystemSaveDataId (and wrappers) + fsDisableAutoSaveDataCreation

Adubbz:
As per title. Needed for reimplementing any sysmodule with savedata.

### DIFF
--- a/nx/include/switch/services/fs.h
+++ b/nx/include/switch/services/fs.h
@@ -82,7 +82,7 @@ typedef struct {
     u64 size;           ///< Size of the save data.
     u64 journalSize;    ///< Journal size of the save data.
     u64 blockSize;      ///< Block size of the save data.
-    u64 ownerId;        ///< Id of the save data's owner.
+    u64 ownerId;        ///< Title Id of the owner of this save data. 0 for SystemSaveData.
     u32 flags;          ///< Save data flags.
     u8 SaveDataSpaceId; ///< See \ref FsSaveDataSpaceId.
     u8 unk;             ///< 0 for SystemSaveData.
--- a/nx/include/switch/services/fs.h
+++ b/nx/include/switch/services/fs.h
@@ -77,6 +77,19 @@ typedef struct
     u64 unk_x38;                    ///< 0 for SystemSaveData/SaveData.
 } FsSave;
 
+/// SaveCreate Struct
+typedef struct
+{
+    u64 journalSize;
+    u64 flags;
+    u64 blockSize;
+    u64 size;
+    u32 unk_x20;
+    u8 SaveDataSpaceId;
+    u8 unk_x25;
+    u8 uninitialized[0x1A];
+} FsSaveCreate;
+
 typedef struct
 {
     u64 saveID_unk;
@@ -226,6 +239,8 @@ Service* fsGetServiceSession(void);
 Result fsOpenBisStorage(FsStorage* out, FsBisStorageId PartitionId);
 Result fsOpenBisFileSystem(FsFileSystem* out, FsBisStorageId PartitionId, const char* string);
 
+Result fsCreateSaveDataFileSystemBySystemSaveDataId(FsSave* save, FsSaveCreate* create);
+
 Result fsIsExFatSupported(bool* out);
 
 /// Do not call this directly, see fs_dev.h.
@@ -245,7 +260,13 @@ Result fsGetRightsIdByPath(const char* path, FsRightsId* out_rights_id);
 
 /// Retrieves the rights id and key generation corresponding to the content path. Only available on [3.0.0+].
 Result fsGetRightsIdAndKeyGenerationByPath(const char* path, u8* out_key_generation, FsRightsId* out_rights_id);
+
+Result fsDisableAutoSaveDataCreation(void);
 // todo: Rest of commands here
+
+// Wrapper(s) for fsCreateSaveDataFileSystemBySystemSaveDataId.
+Result fsCreate_SystemSaveDataWithOwner(FsSaveDataSpaceId space_id, u64 save_data_id, u64 user_id, u64 owner_id, u64 size, u64 journal_size, u32 flags);
+Result fsCreate_SystemSaveData(FsSaveDataSpaceId space_id, u64 save_data_id, u64 size, u64 journal_size, u32 flags);
 
 /// FsFileSystem can be mounted with fs_dev for use with stdio, see fs_dev.h.
 
--- a/nx/include/switch/services/fs.h
+++ b/nx/include/switch/services/fs.h
@@ -78,16 +78,15 @@ typedef struct
 } FsSave;
 
 /// SaveCreate Struct
-typedef struct
-{
-    u64 size;
-    u64 journalSize;
-    u64 blockSize;
-    u64 ownerId;
-    u32 flags;
-    u8 SaveDataSpaceId;
-    u8 unk_x25;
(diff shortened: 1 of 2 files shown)


nx/include/switch/services/fs.h
```diff
@@ -77,6 +77,19 @@ typedef struct
     u64 unk_x38;                    ///< 0 for SystemSaveData/SaveData.
 } FsSave;
 
+/// SaveCreate Struct
+typedef struct
+{
```
yellows8: Move to previous line.

nx/source/services/fs.c
```diff
@@ -633,6 +700,29 @@ Result fsIsExFatSupported(bool* out)
     return rc;
 }
 
+// Wrapper(s) for fsCreateSaveDataFileSystemBySystemSaveDataId.
+Result fsCreate_SystemSaveDataWithOwner(FsSaveDataSpaceId space_id, u64 save_data_id, u64 user_id, u64 owner_id, u64 size, u64 journal_size, u32 flags) {
+    FsSave save;
+    FsSaveCreate create;
+
+    memset(&save, 0, sizeof(save));
+    save.userID = (u128)user_id | ((u128)owner_id << 64);
```
yellows8: Does this really match official sw?(Is this the account userID?)

nx/source/services/fs.c
```diff
@@ -633,6 +700,29 @@ Result fsIsExFatSupported(bool* out)
     return rc;
 }
 
+// Wrapper(s) for fsCreateSaveDataFileSystemBySystemSaveDataId.
+Result fsCreate_SystemSaveDataWithOwner(FsSaveDataSpaceId space_id, u64 save_data_id, u64 user_id, u64 owner_id, u64 size, u64 journal_size, u32 flags) {
```
yellows8: Use same names as the struct fields.

nx/include/switch/services/fs.h
yellows8: Also doc these fields. Did you verify that sizeof(FsSaveCreate) is correct?

yellows8: Update wiki too.

nx/source/services/fs.c
Adubbz: I mistakenly had the size of userid as a u64 in my local db so the two looked like they were getting clobbered together, but I've fixed both now.

nx/source/services/fs.c
Adubbz: Done

nx/include/switch/services/fs.h
Adubbz: Yeah the size should be 0x40 which I verified is correct. Will update the wiki shortly.

Adubbz: Done.

nx/include/switch/services/fs.h
```diff
@@ -77,6 +77,18 @@ typedef struct
     u64 unk_x38;                    ///< 0 for SystemSaveData/SaveData.
 } FsSave;
 
+/// SaveCreate Struct
+typedef struct {
+    u64 size;           ///< Size of the save data.
+    u64 journalSize;    ///< Journal size of the save data.
+    u64 blockSize;      ///< Block size of the save data.
+    u64 ownerId;        ///< Id of the save data's owner.
```
yellows8: Clarify what exactly the Id is?(titleID?)

Adubbz: Yeah it's a title id. Done.

nx/include/switch/services/fs.h
```diff
@@ -226,6 +238,8 @@ Service* fsGetServiceSession(void);
 Result fsOpenBisStorage(FsStorage* out, FsBisStorageId PartitionId);
 Result fsOpenBisFileSystem(FsFileSystem* out, FsBisStorageId PartitionId, const char* string);
 
+Result fsCreateSaveDataFileSystemBySystemSaveDataId(FsSave* save, FsSaveCreate* create);
```
fincs: Const correctness missing for both parameters.

nx/source/services/fs.c
```diff
@@ -633,6 +700,29 @@ Result fsIsExFatSupported(bool* out)
     return rc;
 }
 
+// Wrapper(s) for fsCreateSaveDataFileSystemBySystemSaveDataId.
+Result fsCreate_SystemSaveDataWithOwner(FsSaveDataSpaceId SaveDataSpaceId, u64 saveID, u128 userID, u64 ownerId, u64 size, u64 journalSize, u32 flags) {
+    FsSave save;
+    FsSaveCreate create;
+
+    memset(&save, 0, sizeof(save));
```
fincs: This can be simplified by just doing:
```c
FsSave save = {
    .userID = userID,
    .saveID = saveID,
};
FsSaveCreate create = {
    .size = size,
    .journalSize = journalSize,
    .blockSize = 0x4000,
    .ownerId = ownerId,
    .flags = flags,
    .saveDataSpaceId = saveDataSpaceId,
};
```

nx/include/switch/services/fs.h
```diff
@@ -77,6 +77,18 @@ typedef struct
...
 } FsSave;
 
+/// SaveCreate Struct
+typedef struct {
+    u64 size;           ///< Size of the save data.
+    u64 journalSize;    ///< Journal size of the save data.
+    u64 blockSize;      ///< Block size of the save data.
+    u64 ownerId;        ///< Title id of the owner of this save data. 0 for SystemSaveData.
+    u32 flags;          ///< Save data flags.
+    u8 SaveDataSpaceId; ///< See \ref FsSaveDataSpaceId.
```
fincs: `saveDataSpaceId`

nx/include/switch/services/fs.h
```diff
@@ -245,8 +259,14 @@ Result fsGetRightsIdByPath(const char* path, FsRightsId* out_rights_id);
 
 /// Retrieves the rights id and key generation corresponding to the content path. Only available on [3.0.0+].
 Result fsGetRightsIdAndKeyGenerationByPath(const char* path, u8* out_key_generation, FsRightsId* out_rights_id);
+
+Result fsDisableAutoSaveDataCreation(void);
 // todo: Rest of commands here
 
+// Wrapper(s) for fsCreateSaveDataFileSystemBySystemSaveDataId.
+Result fsCreate_SystemSaveDataWithOwner(FsSaveDataSpaceId SaveDataSpaceId, u64 saveID, u128 userID, u64 ownerId, u64 size, u64 journalSize, u32 flags);
```
fincs: `saveDataSpaceId` (and also in the line below)

nx/include/switch/services/fs.h
Adubbz: I named it like that because the existing FsSave struct had it the same way. Should that be changed too?

fincs: To be honest I find this random capitalization of this one field kind of inconsistent. Pinging @yellows8

Adubbz: On the subject, there seems to be quite a lot of  inconsistencies throughout ``fs.h`` and ``fs.c``. Enum naming is sometimes all caps and other times not, braces are on the next line in some places but on the same line in others, some places use camel casing and others snake casing etc. etc.

yellows8: Yeah change that.

PR #303, opened 2019-07-04
Add hidsysGetUniquePadSerialNumber

friedkeenan:


yellows8:
FWIW controllers' serial#s are also stored in hid-sharedmem, libnx doesn't support it yet though.

PR #304, opened 2019-07-12
Added fsDeleteSaveDataFileSystemBySaveDataSpaceId

Adubbz:
As per title.

### DIFF
--- a/nx/source/services/fs.c
+++ b/nx/source/services/fs.c
@@ -193,9 +193,9 @@ Result fsDeleteSaveDataFileSystemBySaveDataSpaceId(FsSaveDataSpaceId saveDataSpa
     struct {
         u64 magic;
         u64 cmd_id;
-        FsSaveDataSpaceId saveDataSpaceId;
+        u8 saveDataSpaceId;
         u64 saveID;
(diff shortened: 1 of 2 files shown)


nx/source/services/fs.c
```diff
@@ -186,6 +186,42 @@ Result fsCreateSaveDataFileSystemBySystemSaveDataId(const FsSave* save, const Fs
...
 }
 
+Result fsDeleteSaveDataFileSystemBySaveDataSpaceId(FsSaveDataSpaceId saveDataSpaceId, u64 saveID) {
+    IpcCommand c;
+    ipcInitialize(&c);
+
+    struct {
+        u64 magic;
+        u64 cmd_id;
+        FsSaveDataSpaceId saveDataSpaceId;
```
yellows8: Normally this is an u8, is this correct?

Adubbz: It is a u8, however it gets aligned to be 8 bytes. When combined with the save id it should sum to 0x10.

yellows8: Should align properly with u8, and with PACKED removed?

PR #305, opened 2019-07-13
Add a bunch of GetServiceSession functions

friedkeenan:
This is mainly for consistency, since some services have them but most don't. Also it's useful for making IPC calls that aren't in libnx, whether that be for testing them before making a PR to libnx or otherwise. It could also be used to send IPC calls to services that were extended by CFW.

Some services already had a `GetSessionService` or `GetService` function, but I renamed them to `GetServiceSession` to be consistent with the rest of the functions.

There were also some services I didn't add functions for because they have multiple service handles in them:
- applet
- audin
- audout
- nfc
- nifm
- ns
- nv
- pctl
- vi

I'm not sure what to do with them, but I saw that time has multiple service handles and it had a function to get only the main service, so maybe those services should be like time?

There were also a few services that I didn't add functions for because their services aren't stored, but are acquired in each call:
- fatal
- hwopus
- sfdnsres

I figure anyone who needs to send IPC calls to those services can also get the handle themselves beforehand, but I figured I'd mention them anyways.

### DIFF
--- a/nx/include/switch/services/auddev.h
+++ b/nx/include/switch/services/auddev.h
@@ -11,7 +11,7 @@
 
 Result auddevInitialize(void);
 void auddevExit(void);
-Service* auddevgetServiceSession(void);
+Service* auddevGetServiceSession(void);
 
 Result auddevListAudioDeviceName(AudioDeviceName *DeviceNames, s32 max_names, s32 *total_names);
 Result auddevSetAudioDeviceOutputVolume(const AudioDeviceName *DeviceName, float volume);
--- a/nx/include/switch/services/auddev.h
+++ b/nx/include/switch/services/auddev.h
@@ -7,9 +7,11 @@
 #pragma once
 
 #include "../audio/audio.h"
+#include "../services/sm.h"
 
 Result auddevInitialize(void);
 void auddevExit(void);
+Service* auddevgetServiceSession(void);
 
 Result auddevListAudioDeviceName(AudioDeviceName *DeviceNames, s32 max_names, s32 *total_names);
 Result auddevSetAudioDeviceOutputVolume(const AudioDeviceName *DeviceName, float volume);
--- a/nx/source/services/apm.c
+++ b/nx/source/services/apm.c
@@ -41,6 +41,11 @@ void apmExit(void)
     }
 }
 
+Service* apmGetServiceSession(void)
+{
+    return &g_apmSrv;    
+}
+
 static Result _apmGetSession(Service* srv, Service* srv_out, u64 cmd_id) {
     IpcCommand c;
     ipcInitialize(&c);
--- a/nx/source/services/apm.c
+++ b/nx/source/services/apm.c
@@ -41,8 +41,7 @@ void apmExit(void)
     }
 }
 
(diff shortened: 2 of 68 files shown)


nx/include/switch/services/auddev.h
```diff
@@ -7,9 +7,11 @@
 #pragma once
 
 #include "../audio/audio.h"
+#include "../services/sm.h"
 
 Result auddevInitialize(void);
 void auddevExit(void);
+Service* auddevgetServiceSession(void);
```
yellows8: {...}Get{...}

nx/source/services/apm.c
```diff
@@ -41,6 +41,11 @@ void apmExit(void)
     }
 }
 
+Service* apmGetServiceSession(void)
+{
```
yellows8: Move to prev line, for all of these.

nx/include/switch/services/auddev.h
friedkeenan: Fixed

nx/source/services/apm.c
friedkeenan: Should be fixed

nx/source/services/spl.c
```diff
@@ -94,6 +94,11 @@ void splExit(void) {
     return _splSrvExit(&g_splSrv, &g_splRefCnt);
 }
 
+Service* splGetServiceSession(void)
```
SciresM: spl should not return service sessions this way. It will break the muxing we do for < 4.x deprecated service vs >= 4.x modern services.

Use the _splGet*Srv accessors.

friedkeenan: Should those functions be put in the header and renamed to the `GetServiceSession` format?

SciresM: That's kind of awkward due to RsaService, which doesn't have any real service equivalent.

friedkeenan: So what do you suggest? Since I think it'd be awkward having a bunch of functions where all they do is just return what another function returns, and there doesn't seem to be `_splGetSslSrv` or `_splGetManuSrv` functions, which would be awkward for those respective services

SciresM: "where all they do is just return what another function returns" -- I don't think that's awkward.

I would add _splGetSslSrv/_splGetManuSrv, and then add the getters as I described that just forward to the _splGet*Srvs.

Pinging @fincs / @yellows8 about this though.

averne:
I think renamed functions should provide an alias for backward compatibility.

fincs:
@averne We don't usually provide backwards compatibility shims for things that don't get much use.

averne:
Doesn't have to be a shim, you can use the GCC alias attribute, while flagging as deprecated to remove it in the next major revision.

fincs:
Yeah, but there's extra maintenance overhead involved for what is essentially only used in maybe one or two projects.

nx/source/services/spl.c
yellows8: ^sure

friedkeenan: Hopefully I did it correctly with the latest commit

PR #311, opened 2019-07-21
Ccache

fennectech:
Ccache can improve peformance when compiling .  This (shouldent) break anything  and compilation will work properly  on machines without cache


fincs:
This is not the right way to introduce changes to CC/CXX. Please see this file:

https://github.com/devkitPro/buildscripts/blob/master/dka64/rules/base_tools#L28-L29

I'm not sure if it's worth adding ccache support, I'm also not sure if it's possible to have it working across all 3 OSes we support (Windows, Linux, macOS).

PR #314, opened 2019-07-22
Added ability to get raw `UsbDs` structs from `usbComms`.

ischeinkman:
This would allow for non-infinite timeouts on read and write, async IO, and better error handling for users while still allowing us to piggyback off of `usbComms`'s boilerplate reducers. 

Note that I'm not particularly attached to this specific API, so if someone wants to close this and implement it in a different way I'm all for it. Also note that my C is a bit rusty at best, so if there's something off please tell me.

### DIFF
--- a/nx/include/switch/runtime/devices/usb_comms.h
+++ b/nx/include/switch/runtime/devices/usb_comms.h
@@ -40,7 +40,7 @@ size_t usbCommsReadEx(void* buffer, size_t size, u32 interface);
 size_t usbCommsWriteEx(const void* buffer, size_t size, u32 interface);
 
 /// Gets the raw usbDs data for the default usbComms interface.
-Result usbCommsGetInfo(UsbDsInterface ** interface, UsbDsEndpoint ** endpoint_in, UsbDsEndpoint ** endpoint_out);
+Result usbCommsGetInfo(UsbDsInterface ** interface_data, UsbDsEndpoint ** endpoint_in, UsbDsEndpoint ** endpoint_out);
 
 /// Same as usbCommsGetInfo except with the specified interface. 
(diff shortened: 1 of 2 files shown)


yellows8:
See doozer build errors.

fincs:
UsbComms is not really intended to be expanded in any way. Users who need more advanced/flexible USB functionality are advised to call usb:ds APIs directly.

At some point we've considered ripping up the current UsbComms API and starting fresh, not sure if there is enough demand for a better generic USB API.

ischeinkman:
So far I have not seen any project that actually uses the low level `usbDs` API directly; all usb-related projects I am aware of seem to be using `usbComms`. Horizon's `usbDs` API is extremely boilerplate-y, as evidenced by the fact that about half of `usb_comms.c` is devoted to just initialization, so I do think the demand is there (at least as mush as usb-related demand exists at all). Personally, the only features I even particularly care about that I can't do with `usbComms` is async reads and writes through the endpoints' events.

fincs:
> So far I have not seen any project that actually uses the low level usbDs API directly

That's likely due to usbComms requiring less programmer effort vs using usb:ds directly. However, usbComms was never really designed to be anything more than a quick stopgap used to implement nxdbg. Also, if I recall correctly, at some point there was some work in progress code in AMS that used usb:ds directly.

yellows8:
Closing, see previous comments. "... ripping up the current UsbComms API and starting fresh", etc.

SciresM:
> So far I have not seen any project that actually uses the low level `usbDs` API directly; all usb-related projects I am aware of seem to be using `usbComms`. Horizon's `usbDs` API is extremely boilerplate-y, as evidenced by the fact that about half of `usb_comms.c` is devoted to just initialization, so I do think the demand is there (at least as mush as usb-related demand exists at all). Personally, the only features I even particularly care about that I can't do with `usbComms` is async reads and writes through the endpoints' events.

Atmosphere's TMA sysmodule has code that uses usb:ds directly.

PR #315, opened 2019-07-23
Implement various SaveDataExtraData commands

Adubbz:
Another day, another fs PR. As usual, needed for NCM.

### DIFF
--- a/nx/include/switch/services/fs.h
+++ b/nx/include/switch/services/fs.h
@@ -80,13 +80,13 @@ typedef struct
 /// SaveDataExtraData Struct
 typedef struct {
     FsSave save;      ///< Save struct.
-    u64 owner_id;     ///< Title id of the owner of this save data. 0 for SystemSaveData.
+    u64 ownerId;      ///< Title id of the owner of this save data. 0 for SystemSaveData.
     u64 timestamp;    ///< POSIX timestamp.
     u32 flags;        ///< Save data flags. See \ref FsSaveDataFlags.
     u32 unk_x54;      ///< Normally 0. Possibly unused?
-    u64 data_size;    ///< Usable save data size.
-    u64 journal_size; ///< Journal size of the save data.
-    u64 commit_id;    ///< Id of the latest commit.
+    u64 dataSize;     ///< Usable save data size.
+    u64 journalSize;  ///< Journal size of the save data.
+    u64 commitId;     ///< Id of the latest commit.
     u8 unused[0x190]; ///< Uninitialized.
 } FsSaveDataExtraData;
 
@@ -199,8 +199,8 @@ typedef enum
     FsSaveDataType_CacheStorage             = 5, ///< [3.0.0+]
 } FsSaveDataType;
 
-typedef enum
-{
+/// SaveDataFlags
+typedef enum {
     FsSaveDataFlags_SurviveFactoryReset                    = BIT(0),
     FsSaveDataFlags_SurviveFactoryResetForRefurbishment    = BIT(1),
     FsSaveDataFlags_SurviveFactoryResetWithoutUserSaveData = BIT(2),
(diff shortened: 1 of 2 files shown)


nx/include/switch/services/fs.h
```diff
@@ -77,13 +77,26 @@ typedef struct
     u64 unk_x38;                    ///< 0 for SystemSaveData/SaveData.
 } FsSave;
 
+/// SaveDataExtraData Struct
+typedef struct {
+    FsSave save;      ///< Save struct.
+    u64 owner_id;     ///< Title id of the owner of this save data. 0 for SystemSaveData.
```
yellows8: ownerId

nx/include/switch/services/fs.h
```diff
@@ -77,13 +77,26 @@ typedef struct
...
+    u64 timestamp;    ///< POSIX timestamp.
+    u32 flags;        ///< Save data flags. See \ref FsSaveDataFlags.
+    u32 unk_x54;      ///< Normally 0. Possibly unused?
+    u64 data_size;    ///< Usable save data size.
```
yellows8: dataSize

nx/include/switch/services/fs.h
```diff
@@ -77,13 +77,26 @@ typedef struct
...
+    u64 journal_size; ///< Journal size of the save data.
```
yellows8: journalSize

nx/include/switch/services/fs.h
```diff
@@ -77,13 +77,26 @@ typedef struct
...
+    u64 commit_id;    ///< Id of the latest commit.
```
yellows8: commitId

nx/include/switch/services/fs.h
```diff
@@ -186,6 +199,13 @@ typedef enum
     FsSaveDataType_CacheStorage             = 5, ///< [3.0.0+]
 } FsSaveDataType;
 
+typedef enum
+{
```
yellows8: Move to previous line.

nx/include/switch/services/fs.h
```diff
@@ -186,6 +199,13 @@ typedef enum
     FsSaveDataType_CacheStorage             = 5, ///< [3.0.0+]
 } FsSaveDataType;
 
+typedef enum
```
yellows8: "/// SaveDataFlags" before this, or similar.

nx/include/switch/services/fs.h
```diff
@@ -249,6 +269,9 @@ Result fsDeleteSaveDataFileSystemBySaveDataSpaceId(FsSaveDataSpaceId saveDataSpa
 
 Result fsIsExFatSupported(bool* out);
 Result fsOpenGameCardFileSystem(FsFileSystem* out, const FsGameCardHandle* handle, FsGameCardPartiton partition);
+Result fsReadSaveDataFileSystemExtraDataBySaveDataSpaceId(void* buf, size_t len, FsSaveDataSpaceId saveDataSpaceId, u64 saveID);
```
yellows8: Shouldn't these use the above struct?

Adubbz: I was hesitant to do that because it implies the commands only accept a fixed size struct. That may not necessarily always be the case, especially given the large amount of uninitialized data it contains at the end. N may potentially use buffers less than the full size, so I wanted to keep the option there to preserve that behaviour (should anything use it).

yellows8: ok

nx/include/switch/services/fs.h
fincs: Actually: https://github.com/switchbrew/libnx/blob/master/CODESTYLE.md#struct-members

nx/include/switch/services/fs.h
fincs: Same thing here (see codestyle.md)

nx/include/switch/services/fs.h
Adubbz: Well, the other thing is that most of the structs in fs.h don’t use that style (and it’s a similar case in other headers too). Kinda hard to tell which one to use tbh.

PR #316, opened 2019-07-24
Added viDestroyManagedLayer

averne:


nx/source/services/vi.c
```diff
@@ -960,7 +992,7 @@ Result viCloseLayer(ViLayer *layer)
         } *resp = r.Raw;
 
         rc = resp->result;
-        memset(layer, 0, sizeof(ViLayer));
+        layer->initialized = false;
```
fincs: Why was this changed?

averne: To close a managed layer you need to call viCloseLayer and viDestroyManagedLayer successively, but since the layer id was zeroed you'd need to cache it otherwise.

fincs: This tells me there's a design problem with the API. I don't think it's reasonable to call a method on an object that has been closed. Probably, viDestroyManagedLayer should be internally called somewhere within viCloseLayer.

averne: Ok nevermind, for some reason I remembered calling viDestroyManagedLayer without calling viCloseLayer first would make it error, but after testing it appears that not.
I guess I should zero the layer similar to viCloseLayer then?

fincs: Sure.

PR #318, opened 2019-08-07
vi: add ViPowerState_On_Deprecated enum entry

SciresM:


nx/include/switch/services/vi.h
```diff
@@ -45,9 +45,11 @@ typedef enum {
 
 /// Used with viSetDisplayPowerState.
 typedef enum {
-    ViPowerState_Off = 0,
-    ViPowerState_NotScanning = 1,
-    ViPowerState_On = 2,
+    ViPowerState_Off           = 0, ///< Screen is off.
+    ViPowerState_NotScanning   = 1, ///< Screen is on, but not scanning content.
```
yellows8: Add [3.0.0+] for these two?

PR #320, opened 2019-08-14
Add fsdevDeleteDevice function

XorTroll:
This way one can unmount filesystems without closing them (in case they would be used later without fsdev/stdio)


fincs:
Instead of adding a second function to unmount a device, I would instead make FsFileSystem ownership part of the device state, and add a way to mount a filesystem without transferring its ownership to fsdev.

yellows8:
Pinging @XorTroll ^.

fincs:
Pinging @XorTroll again

XorTroll:
Oh well, after checking on my project where I got this feature from, it turns out that it was actually unused :P
Anyway, if you guys want to implement it in other way, ignore this PR

fincs:
Okay.

PR #321, opened 2019-08-17
Add some FS commands

Thealexbarney:
fsOpenCustomStorageFileSystem
fsExtendSaveDataFileSystem
fsSetGlobalAccessLogMode
fsGetGlobalAccessLogMode

### DIFF
--- a/nx/include/switch/services/fs.h
+++ b/nx/include/switch/services/fs.h
@@ -84,16 +84,16 @@ typedef struct {
     u64 timestamp;    ///< POSIX timestamp.
     u32 flags;        ///< Save data flags. See \ref FsSaveDataFlags.
     u32 unk_x54;      ///< Normally 0. Possibly unused?
-    u64 dataSize;     ///< Usable save data size.
-    u64 journalSize;  ///< Journal size of the save data.
+    s64 dataSize;     ///< Usable save data size.
+    s64 journalSize;  ///< Journal size of the save data.
     u64 commitId;     ///< Id of the latest commit.
     u8 unused[0x190]; ///< Uninitialized.
 } FsSaveDataExtraData;
 
 /// SaveCreate Struct
 typedef struct {
-    u64 size;           ///< Size of the save data.
-    u64 journalSize;    ///< Journal size of the save data.
+    s64 size;           ///< Size of the save data.
+    s64 journalSize;    ///< Journal size of the save data.
     u64 blockSize;      ///< Block size of the save data.
     u64 ownerId;        ///< Title id of the owner of this save data. 0 for SystemSaveData.
     u32 flags;          ///< Save data flags. See \ref FsSaveDataFlags.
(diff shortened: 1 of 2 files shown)


nx/source/services/fs.c
```diff
@@ -891,6 +932,123 @@ Result fsWriteSaveDataFileSystemExtraData(const void* buf, size_t len, FsSaveDat
...
+        resp = r.Raw;
+
+        rc = resp->result;
+    }
+
+    return rc;
+}
+
+Result fsSetGlobalAccessLogMode(u32 mode)
+{
```
yellows8: Move to previous line, same below.

nx/include/switch/services/fs.h
```diff
@@ -272,6 +278,7 @@ Result fsOpenGameCardFileSystem(FsFileSystem* out, const FsGameCardHandle* handl
 Result fsReadSaveDataFileSystemExtraDataBySaveDataSpaceId(void* buf, size_t len, FsSaveDataSpaceId saveDataSpaceId, u64 saveID);
 Result fsReadSaveDataFileSystemExtraData(void* buf, size_t len, u64 saveID);
 Result fsWriteSaveDataFileSystemExtraData(const void* buf, size_t len, FsSaveDataSpaceId saveDataSpaceId, u64 saveID);
+Result fsExtendSaveDataFileSystem(FsSaveDataSpaceId saveDataSpaceId, u64 saveID, s64 dataSize, s64 journalSize);
```
yellows8: Add comments for required sysver, same below.

yellows8:
Can you doc what CustomStorage is on wiki?

yellows8:
That wiki edit doesn't explain what "CustomStorage" actually is (as in more than just the enum).

Thealexbarney:
I hadn't finished adding the rest yet. The other part is added now

nx/include/switch/services/fs.h
```diff
@@ -272,6 +278,7 @@ Result fsOpenGameCardFileSystem(FsFileSystem* out, const FsGameCardHandle* handl
 Result fsReadSaveDataFileSystemExtraDataBySaveDataSpaceId(void* buf, size_t len, FsSaveDataSpaceId saveDataSpaceId, u64 saveID);
 Result fsReadSaveDataFileSystemExtraData(void* buf, size_t len, u64 saveID);
 Result fsWriteSaveDataFileSystemExtraData(const void* buf, size_t len, FsSaveDataSpaceId saveDataSpaceId, u64 saveID);
+Result fsExtendSaveDataFileSystem(FsSaveDataSpaceId saveDataSpaceId, u64 saveID, s64 dataSize, s64 journalSize); /// 3.0.0+
```
yellows8: Is it known whether dataSize/journalSize used elsewhere are also signed?

Thealexbarney: Yes. The fields in the save create struct also use signed division

yellows8: Update those then?

Thealexbarney: I didn't check FsSaveDataExtraData, but I'm assuming it's the same there

PR #324, opened 2019-09-02
Require whole mount name to match when finding a romfs

friedkeenan:
Before, only the first letters of the mount name had to match, so that if you had a romfs mounted at `romfs` and tried to unmount `romfshhhhhh` it would still unmount `romfs`

### DIFF
--- a/nx/source/runtime/devices/romfs_dev.c
+++ b/nx/source/runtime/devices/romfs_dev.c
@@ -170,7 +170,7 @@ static romfs_mount *romfsFindMount(const char *name)
         }
         else if(mount->setup) //Find the mount with the input name.
         {
-            if(strncmp(mount->name, name, strlen(mount->name))==0)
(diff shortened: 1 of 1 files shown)


nx/source/runtime/devices/romfs_dev.c
```diff
@@ -170,7 +170,7 @@ static romfs_mount *romfsFindMount(const char *name)
         }
         else if(mount->setup) //Find the mount with the input name.
         {
-            if(strncmp(mount->name, name, strlen(mount->name))==0)
+            if(strncmp(mount->name, name, sizeof(mount->name)-1)==0)
```
fincs: Don't do `-1`, let strncmp make sure that the last character before the limit is forcibly the null terminator.

friedkeenan: Done

PR #332, opened 2019-09-30
ncm: Use u8 rather than actual types to ensure correct struct sizes

Adubbz:
As per title. Noticed this was missed in my previous changes.


fincs:
Quick question before merging right away: are these structs used during IPC?

Adubbz:
``NcmContentMetaKey`` is directly accepted used an input/output, ``NcmContentInfo``/``NcmContentMetaHeader`` are part of buffer outputs.

fincs:
Mkay, merged.

PR #335, opened 2019-10-04
Add psel (playerSelect) support

XorTroll:
Everything was tested working fine (all modes and flags)

### DIFF
--- a/nx/include/switch/applets/psel.h
+++ b/nx/include/switch/applets/psel.h
@@ -54,7 +54,7 @@ Result pselUiCreate(PselUiSettings *ui, PselUiMode mode);
  * @param ui PselUiSettings struct.
  * @param user_id user ID.
  */
-void pselUiAddInvalidUser(PselUiSettings *ui, AccountUid *user_id);
+void pselUiAddUser(PselUiSettings *ui, AccountUid *user_id);
 
 /**
  * @brief Sets whether users can be created in the applet
@@ -87,9 +87,33 @@ NX_CONSTEXPR void pselUiSetSkipEnabled(PselUiSettings *ui, bool flag) {
 }
 
 /**
- * @brief Shows the applet with the specified UI settings.
+ * @brief Shows playerSelect applet with the specified UI settings.
  * @param ui PselUiSettings struct.
- * @param out_uid Selected user ID.
+ * @param out_user Selected user ID.
  * @note If user skips (see \ref pselUiSetSkipEnabled) this will return successfully but the output ID will be 0.
  */
 Result pselUiShow(PselUiSettings *ui, AccountUid *out_user);
+
+/**
+ * @brief Shows playerSelect applet to select a user.
+ * @param out_user Returned selected user ID.
+ */
+Result pselShowUserSelector(AccountUid *out_user);
+
+/**
+ * @brief Shows playerSelect applet to create a user.
+ * @param out_user Returned created user ID.
+ */
+Result pselShowUserCreator(AccountUid *out_user);
+
+/**
+ * @brief Shows playerSelect applet to change a user's icon.
+ * @param user Input user ID.
+ */
+Result pselShowIconEditor(AccountUid *user);
+
+/**
+ * @brief Shows playerSelect applet to change a user's nickname.
+ * @param user Input user ID.
+ */
+Result pselShowNicknameEditor(AccountUid *user);
--- a/nx/include/switch/applets/psel.h
+++ b/nx/include/switch/applets/psel.h
@@ -54,7 +54,7 @@ Result pselUiCreate(PselUiSettings *ui, PselUiMode mode);
  * @param ui PselUiSettings struct.
  * @param user_id user ID.
  */
-void pselUiAddInvalidUser(PselUiSettings *ui, AccountUid user_id);
+void pselUiAddInvalidUser(PselUiSettings *ui, AccountUid *user_id);
 
 /**
  * @brief Sets whether users can be created in the applet
--- a/nx/include/switch/applets/psel.h
+++ b/nx/include/switch/applets/psel.h
@@ -62,21 +62,29 @@ void pselUiAddInvalidUser(PselUiSettings *ui, AccountUid *user_id);
  * @param flag Flag value.
  * @note Only used for ::PselUiMode_SelectUser
  */
-void pselUiSetAllowUserCreation(PselUiSettings *ui, bool flag);
+NX_CONSTEXPR void pselUiSetAllowUserCreation(PselUiSettings *ui, bool flag) {
+    if(ui->mode == PselUiMode_SelectUser) {
+        ui->allowUserCreation = flag;
+    }
+}
 
 /**
  * @brief Sets whether users need to be linked to a Nintendo account.
  * @param ui PselUiSettings struct.
  * @param flag Flag value.
  */
-void pselUiSetNetworkServiceRequired(PselUiSettings *ui, bool flag);
+NX_CONSTEXPR void pselUiSetNetworkServiceRequired(PselUiSettings *ui, bool flag) {
+    ui->networkServiceRequired = flag;
+}
 
 /**
  * @brief Sets whether selection can be skipped (with a new button)
  * @param ui PselUiSettings struct.
  * @param flag Flag value.
  */
-void pselUiSetSkipEnabled(PselUiSettings *ui, bool flag);
+NX_CONSTEXPR void pselUiSetSkipEnabled(PselUiSettings *ui, bool flag) {
+    ui->skipEnabled = flag;
+}
 
 /**
  * @brief Shows the applet with the specified UI settings.
--- a/nx/include/switch/applets/psel.h
+++ b/nx/include/switch/applets/psel.h
@@ -25,21 +25,21 @@ typedef enum {
 
 /// UI settings for playerSelect.
 typedef struct {
-    u32 mode;                                                   ///< UI mode, see \ref PselUiMode.
-    u32 dialogType;                                             ///< Dialog type
-    union { u128 invalidUserList[ACC_USER_LIST_SIZE]; } PACKED; ///< List of user IDs which will be disabled.
-    u8 unk_x0[0x8];                                             ///< Unknown.
-    u8 networkServiceRequired;                                  ///< Whether the user needs to be linked to a Nintendo account.
-    u8 unk_x1[0x2];                                             ///< Unknown.
-    u8 allowUserCreation;                                       ///< (With ::PselUiMode_SelectUser) enables the option to create a new user.
-    u8 skipEnabled;                                             ///< Enables the option to skip user selection (a new button is shown)
-    u8 unk_x2[0xB];                                             ///< Unknown.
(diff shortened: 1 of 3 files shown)


yellows8:
Can you update this so that it matches sdknso better? See also pctlauth.

Guess this doesn't handle newer versions either, 9.0 has a newer commonargs-version. There's also an option for pushing an additional storage, so libappletLaunch can't be used for this.

XorTroll:
What exactly do you mean with "match sdknso better"?
To have similar enum names, functions, descriptions...?
I'm looking on version values used (sdk uses 0x10000 but not sure for which version)

XorTroll:
A newer sdk nso uses 0x20000 as version, so I'm assuming 9.0+ is 0x20000.

yellows8:
Names + funcs. See friendsLa, sdknso handling seems to be similar-ish.

XorTroll:
Still need to figure out for what is the larger storage used for (all the added UI modes don't seem to use it...)

yellows8:
Yeah that LargeStorage doesn't seem to be used in sdksno (dunno if qlaunch etc uses that func?).

Also, if any of the impl'd functionality needs a certain sysver, there should be incompat-sysver checks for it.

nx/include/switch/applets/psel.h
```diff
@@ -0,0 +1,87 @@
...
+ * @brief Wrapper for using playerSelect (user selection applet).
+ * @author XorTroll
+ * @copyright libnx Authors
+ */
+#pragma once
+#include "../types.h"
+#include "../services/applet.h"
+#include "../services/acc.h"
+
+// playerSelect UI modes.
```
yellows8: ///

nx/include/switch/applets/psel.h
```diff
@@ -0,0 +1,87 @@
...
+    PselUiMode_LicenseRequirementsForNetworkService = 10, ///< LicenseRequirementsForNetworkService
+    PselUiMode_NaLoginTest                          = 12, ///< NaLoginTest
+} PselUiMode;
+
+// UI settings for playerSelect.
+typedef struct {
+    u32 mode;                                     ///< UI mode, see \ref PselUiMode.
+    u32 dialogType;                               ///< Dialog type
+    u128 invalidUserList[ACC_USER_LIST_SIZE];     ///< List of user IDs which will be disabled.
+    u8 unk1[8];                                   ///< Unknown.
```
yellows8: `u8 unk_x{offset}[{hex}];`

nx/include/switch/applets/psel.h
```diff
@@ -0,0 +1,87 @@
...
+    u8 networkServiceRequired;                    ///< Whether the user needs to be linked to a Nintendo account.
+    u8 unk2[2];                                   ///< Unknown.
+    u8 allowUserCreation;                         ///< (With ::PselUiMode_SelectUser) enables the option to create a new user.
+    u8 skipEnabled;                               ///< Enables the option to skip user selection (a new button is shown)
+    u8 unk3[11];                                  ///< Unknown.
+} PselUiSettings;
+
+// Result data sent after execution.
+typedef struct {
+    u32 result;                      ///< Result code.
```
yellows8: Can use Result for this.

nx/include/switch/applets/psel.h
```diff
@@ -0,0 +1,87 @@
...
+    u128 userId;                     ///< Selected user ID.
```
yellows8: u128 is aligned to 0x10-bytes, not 8-bytes. userIDs (including this one) need the latter, see the handling elsewhere for that.

nx/source/applets/psel.c
```diff
@@ -0,0 +1,66 @@
+#include <string.h>
+#include "types.h"
+#include "result.h"
+#include "services/applet.h"
+#include "services/set.h"
```
yellows8: Guess this was supposed to be acc.h?

nx/source/applets/psel.c
```diff
@@ -0,0 +1,66 @@
...
+    LibAppletArgs args;
+    u32 la_ver = _pselGetLaVersion();
+    libappletArgsCreate(&args, la_ver);
+    PselResult res;
+    size_t reply_size;
+
+    rc = libappletLaunch(AppletId_playerSelect, &args, ui, sizeof(PselUiSettings), &res, sizeof(res), &reply_size);
+    
+    if (R_SUCCEEDED(rc)) {
+        if (R_FAILED(res.result)) rc = MAKERESULT(Module_Libnx, LibnxError_LibAppletBadExit);
```
yellows8: `rc = res.result;`

nx/include/switch/applets/psel.h
```diff
@@ -0,0 +1,87 @@
...
+    PselUiMode_NicknameEditor                       = 4,  ///< NicknameEditor
+    PselUiMode_ForStarter                           = 5,  ///< Mode "starter" uses to register the console's first user on the initial setup
+    PselUiMode_NetworkServiceAccountRegistration    = 8,  ///< NetworkServiceAccountRegistration
+    PselUiMode_NintendoAccountNnidLinker            = 9,  ///< NintendoAccountNnidLinker
+    PselUiMode_LicenseRequirementsForNetworkService = 10, ///< LicenseRequirementsForNetworkService
+    PselUiMode_NaLoginTest                          = 12, ///< NaLoginTest
+} PselUiMode;
+
+// UI settings for playerSelect.
+typedef struct {
```
yellows8: Official sw stores structs within UiSettings, update this struct for that (one of those funcs also writes a titleID to UiSettings, FWIW).

nx/source/applets/psel.c
XorTroll: The result returned on the out data isn't actually a normal result I think, according to switchbrew
https://switchbrew.org/wiki/Profile_Selector

nx/include/switch/applets/psel.h
XorTroll: As commented below, according to switchbrew the result returned on the out data isn't actually a normal result

nx/source/applets/psel.c
yellows8: Then don't use R_FAILED?

XorTroll: Well on this sense it is also 0 on success and other value on failure it so the macro was suitable, if this leads to confusion won't use it I guess

nx/include/switch/applets/psel.h
```diff
@@ -0,0 +1,87 @@
...
+    PselUiMode_NintendoAccountNnidLinker            = 9,  ///< NintendoAccountNnidLinker
+    PselUiMode_LicenseRequirementsForNetworkService = 10, ///< LicenseRequirementsForNetworkService
+    PselUiMode_NaLoginTest                          = 12, ///< NaLoginTest
+} PselUiMode;
+
+/// UI settings for playerSelect.
+typedef struct {
+    u32 mode;                                                   ///< UI mode, see \ref PselUiMode.
+    u32 dialogType;                                             ///< Dialog type
+    union { u128 invalidUserList[ACC_USER_LIST_SIZE]; } PACKED; ///< List of user IDs which will be disabled.
```
fincs: Use AccountUid here too.

nx/include/switch/applets/psel.h
```diff
@@ -0,0 +1,87 @@
...
+    u8 unk_x1[0x2];                                             ///< Unknown.
+    u8 allowUserCreation;                                       ///< (With ::PselUiMode_SelectUser) enables the option to create a new user.
+    u8 skipEnabled;                                             ///< Enables the option to skip user selection (a new button is shown)
+    u8 unk_x2[0xB];                                             ///< Unknown.
+} PselUiSettings;
+
+/// Result data sent after execution.
+typedef struct {
+    u32 result;                      ///< Result code.
+    union { u128 userId; } PACKED;   ///< Selected user ID.
```
fincs: Use AccountUid here too.

nx/include/switch/applets/psel.h
```diff
@@ -0,0 +1,87 @@
...
+ */
+void pselUiSetSkipEnabled(PselUiSettings *ui, bool flag);
+
+/**
+ * @brief Shows the applet with the specified UI settings.
+ * @param ui PselUiSettings struct.
+ * @param out_uid Selected user ID.
+ * @note If user skips (see \ref pselUiSetSkipEnabled) this will return successfully but the output ID will be 0.
+ */
+Result pselUiShow(PselUiSettings *ui, u128 *out_user);
```
fincs: Use AccountUid here too.

nx/include/switch/applets/psel.h
XorTroll: Ah, didn't know it has a custom type now

XorTroll:
On another topic, sdkver 5.3.0 uses 0x10000 LaVersion, while 7.3.2 uses 0x20000 (unsure which console versions they are paired to)

yellows8:
See the Doozer build log.

XorTroll:
I haven't updated libnx for a while in my fork, is it the reason of that fail?

yellows8:
yes

fincs:
PR needs rebasing against libnx master.

XorTroll:
Sdknso 8.2.0 also uses 0x20000 version, btw

nx/include/switch/applets/psel.h
XorTroll: What exactly do you mean with "stores structs within UiSettings", that its content is split into different structs?

yellows8: Yes, same as official sw.

nx/source/applets/psel.c
```diff
@@ -0,0 +1,66 @@
...
+    for(i = 0; i < ACC_USER_LIST_SIZE; i++) {
+
+        if(!accountUidIsValid(&ui->invalidUserList[i])) {
+            __builtin_memcpy(&ui->invalidUserList[i], user_id, sizeof(AccountUid));
+            break;
+        }
+    }
+}
+
+void pselUiSetAllowUserCreation(PselUiSettings *ui, bool flag) {
```
fincs: Should be a NX_CONSTEXPR function in the header

nx/source/applets/psel.c
```diff
@@ -0,0 +1,66 @@
...
+    if(ui->mode == PselUiMode_SelectUser) {
+        ui->allowUserCreation = flag;
+    }
+}
+
+void pselUiSetNetworkServiceRequired(PselUiSettings *ui, bool flag) {
```
fincs: Should be a NX_CONSTEXPR function in the header

nx/source/applets/psel.c
```diff
@@ -0,0 +1,66 @@
...
+    ui->networkServiceRequired = flag;
+}
+
+void pselUiSetSkipEnabled(PselUiSettings *ui, bool flag) {
```
fincs: Should be a NX_CONSTEXPR function

yellows8:
An API that's semi-similar to official sw should also be used, like the rest of the library-applets. For example: `{prefix}ShowUserIconEditor(AccountUid uid)`

XorTroll:
Added a simplified API on the latest commits

yellows8:
Might adjust this later (?).

PR #345, opened 2019-11-07
hiddbg: Add a function to get Hdls' TransferMemory

cathery:
When low memory usage is important, functions like `hiddbgDumpHdlsStates()` and `hiddbgDumpHdlsNpadAssignmentState()` won't be efficient as they copy the entire struct.
Since there is no need to copy the struct at all, the function `hiddbgGetWorkBufferTransferMemoryAddress()` lets the user read from transfer memory directly.

### DIFF
--- a/nx/source/services/hiddbg.c
+++ b/nx/source/services/hiddbg.c
@@ -392,8 +392,7 @@ Result hiddbgReleaseHdlsWorkBuffer(void) {
     return rc;
 }
 
-Result hiddbgIsHdlsVirtualDeviceAttached(u64 HdlsHandle, bool *isAttached)
-{
+Result hiddbgIsHdlsVirtualDeviceAttached(u64 HdlsHandle, bool *isAttached) {
     Result rc = 0;
 
     if (hosversionBefore(7,0,0))
--- a/nx/source/services/hiddbg.c
+++ b/nx/source/services/hiddbg.c
@@ -403,30 +403,22 @@ Result hiddbgIsHdlsVirtualDeviceAttached(u64 HdlsHandle, bool *isAttached)
         return MAKERESULT(Module_Libnx, LibnxError_NotInitialized);
 
     rc = _hiddbgCmdNoIO(327);
-    if (R_FAILED(rc))
-        return rc;
-    if (isAttached)
-    {
+    if (R_FAILED(rc)) return rc;
+    if (isAttached) {
 		*isAttached = false;
-        if (hosversionBefore(9, 0, 0))
-        {
+        if (hosversionBefore(9, 0, 0)) {
             HiddbgHdlsStateListV7 *stateList = (HiddbgHdlsStateListV7 *)(g_hiddbgHdlsTmem.src_addr);
-            for (s32 i = 0; i < stateList->total_entries; i++)
-            {
-                if (stateList->entries[i].HdlsHandle == HdlsHandle)
-                {
+            for (s32 i = 0; i < stateList->total_entries; i++) {
+                if (stateList->entries[i].HdlsHandle == HdlsHandle) {
                     *isAttached = true;
                     break;
                 }
             }
         }
-        else
-        {
+        else {
             HiddbgHdlsStateList *stateList = (HiddbgHdlsStateList *)(g_hiddbgHdlsTmem.src_addr);
-            for (s32 i = 0; i < stateList->total_entries; i++)
-            {
-                if (stateList->entries[i].HdlsHandle == HdlsHandle)
-                {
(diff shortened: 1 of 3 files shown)


fincs:
I think it would be a better idea to either a) return the address of the work buffer directly instead of the tmem object (and name it `hiddbgGetWorkBuffer`), or b) add a special function to detect if a virtual device is attached, which is what you actually want to use this for (name it `hiddbgIsHdlsVirtualDeviceAttached`)

yellows8:
`hiddbgGetWorkBuffer` <- Name it `hiddbgGetHdlsWorkBuffer`. That doesn't resolve {directly using ipc cmd which libnx already implements} though, so b) is preferred.

cathery:
I'll implement b). I don't know why I didn't think of that

nx/source/services/hiddbg.c
```diff
@@ -392,6 +392,42 @@ Result hiddbgReleaseHdlsWorkBuffer(void) {
     return rc;
 }
 
+Result hiddbgIsHdlsVirtualDeviceAttached(u64 HdlsHandle, bool *isAttached)
+{
+    Result rc = 0;
+
+    if (hosversionBefore(7, 0, 0))
```
yellows8: Remove spaces after "," for hos-versions.

nx/source/services/hiddbg.c
```diff
@@ -392,6 +392,42 @@ Result hiddbgReleaseHdlsWorkBuffer(void) {
...
+
+    if (!g_hiddbgHdlsInitialized)
+        return MAKERESULT(Module_Libnx, LibnxError_NotInitialized);
+
+    rc = _hiddbgCmdNoIO(327);
+    if (R_FAILED(rc)) return rc;
+    if (isAttached) {
+		*isAttached = false;
+        if (hosversionBefore(9, 0, 0)) {
+            HiddbgHdlsStateListV7 *stateList = (HiddbgHdlsStateListV7 *)(g_hiddbgHdlsTmem.src_addr);
```
yellows8: Remove space for the HiddbgHdlsStateListV7 cast, same below.

nx/source/services/hiddbg.c
```diff
@@ -392,6 +392,42 @@ Result hiddbgReleaseHdlsWorkBuffer(void) {
...
+    if (hosversionBefore(7, 0, 0))
+        return MAKERESULT(Module_Libnx, LibnxError_IncompatSysVer);
+
+    if (!g_hiddbgHdlsInitialized)
+        return MAKERESULT(Module_Libnx, LibnxError_NotInitialized);
+
+    rc = _hiddbgCmdNoIO(327);
+    if (R_FAILED(rc)) return rc;
+    if (isAttached) {
+		*isAttached = false;
```
yellows8: tabs->spaces

nx/source/services/hiddbg.c
```diff
@@ -392,6 +392,42 @@ Result hiddbgReleaseHdlsWorkBuffer(void) {
...
+        if (hosversionBefore(9, 0, 0)) {
+            HiddbgHdlsStateListV7 *stateList = (HiddbgHdlsStateListV7 *)(g_hiddbgHdlsTmem.src_addr);
+            for (s32 i = 0; i < stateList->total_entries; i++) {
```
yellows8: Use same style with `for` as rest of hiddbg, same below.

nx/source/services/hiddbg.c
```diff
@@ -392,6 +392,42 @@ Result hiddbgReleaseHdlsWorkBuffer(void) {
     return rc;
 }
 
+Result hiddbgIsHdlsVirtualDeviceAttached(u64 HdlsHandle, bool *isAttached)
+{
```
yellows8: Move to previous line.

nx/source/services/hiddbg.c
```diff
@@ -392,6 +392,41 @@ Result hiddbgReleaseHdlsWorkBuffer(void) {
...
+    if (!g_hiddbgHdlsInitialized)
+        return MAKERESULT(Module_Libnx, LibnxError_NotInitialized);
+
+    rc = _hiddbgCmdNoIO(327);
+    if (R_FAILED(rc)) return rc;
+    if (isAttached) {
+        *isAttached = false;
+        if (hosversionBefore(9,0,0)) {
+            HiddbgHdlsStateListV7 *stateList = (HiddbgHdlsStateListV7*)(g_hiddbgHdlsTmem.src_addr);
+            for (s32 i=0; i <stateList->total_entries; i++) {
```
yellows8: Remove the space for `i <`, same below.

PR #349, opened 2019-11-21
Add FsSaveDataFlag for BIT(3)

shchmue:
ref https://github.com/Thealexbarney/LibHac/blob/f3b5cad94b1d0ce5a1799048172f99e5fd7c3f59/src/LibHac/FsService/FileSystemProxy.cs#L178

in testing, present in the following saves in 9.0.1:
```
save id              flags owner id
8000000000000010     1100  010000000000001e
8000000000000050     1000  0100000000000009
80000000000000a0     1000  010000000000000c
80000000000000a1     1000  010000000000000c
80000000000000a2     1000  010000000000000c
8000000000000190     1000  0100000000000031
```
makes sense for account, settings, prepo saves (not sure why glue)


SciresM:
Probably needs a comment about what firmware version introduced this bit.

shchmue:
I haven't looked at FS myself, so all I know is there are saves with this bit set in my 8.1.0 and 9.0.1 backups but not in 4.1.0

SciresM:
It's pretty recent, I guess. Look up where the code that uses it was introduced?

PR #350, opened 2019-12-03
rwlock: revamp completely

SciresM:
Implements rwlockTryReadLock and rwlockTryWriteLock.
Also implements rwlockIsWriteLockHeldByCurrentThread and rwlockIsOwnedByCurrentThread.

Also re-designs RwLock to have semantics identical toNintendo's (nn::os::ReaderWriterLock). 

The upshot is mostly that the lock is now fully recursive/write-preferring.

### DIFF
--- a/nx/source/kernel/rwlock.c
+++ b/nx/source/kernel/rwlock.c
@@ -54,7 +54,7 @@ bool rwlockTryReadLock(RwLock* r) {
 
     const bool got_lock = r->write_waiter_count == 0;
     if (got_lock) {
-        r->read_waiter_count++;
+        r->read_lock_count++;
     }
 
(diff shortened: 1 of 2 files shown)


nx/source/kernel/rwlock.c
```diff
@@ -1,59 +1,170 @@
...
+        return true;
+    }
+
+    if (!mutexTryLock(&r->mutex)) {
+        return false;
+    }
+
+    const bool got_lock = r->write_waiter_count == 0;
+    if (got_lock) {
+        r->read_waiter_count++;
```
fincs: This accomplishes nothing and there's nothing decrementing the waiter count, since this thread isn't put to sleep. Probably this if section should be removed.

PR #354, opened 2019-12-12
hidKeysAll- and hidGetControllerCount()

SunTheCourier:
Create functions to get keys pressed, held, and up across all controllers aswell as get total connected controller count. key functions similarly function like most official applications.


yellows8:
These don't belong in libnx, it's essentially "wrappers" for hidKeys*()/hidIsControllerConnected() - apps should do it themselves if that's wanted. Not sure what the use-case for hidGetControllerCount() is either.

fincs:
There's a HID refactor/rewrite/redesign planned at some point that will fix numerous issues with the current HID code.

PR #356, opened 2019-12-15
Add set:cal service with a few cal functions.

ITotalJustice:
Hi, using [switchbrew docs](https://switchbrew.org/wiki/Settings_services#set:cal) I added the set:cal service and a few of the functions that were documented.

I have tested them and they work.

Let me know if you would like me to rename anything or move the functions around.

### DIFF
--- a/nx/include/switch/services/set.h
+++ b/nx/include/switch/services/set.h
@@ -271,6 +271,36 @@ void setExit(void);
 /// Gets the Service object for the actual set service session.
 Service* setGetServiceSession(void);
 
+/// Initialize setcal.
+Result setcalInitialize(void);
+
+/// Exit setcal.
+void setcalExit(void);
+
+/// Gets the Service object for the actual setcal service session.
+Service* setcalGetServiceSession(void);
+
+/// Gets EciDeviceCertificate of size 0x180.
+Result setcalGetEciDeviceCertificate(void *certificate, size_t size);
+
+/// Gets EticketDeviceCertificate of size 0x240.
+Result setcalGetEticketDeviceCertificate(void *certificate, size_t size);
+
+/// Gets SslKey of size 0x134.
+Result setcalGetSslKey(void *key, size_t size);
+
+/// Gets SslCertificate of size 0x804.
+Result setcalGetSslCertificate(void *certificate, size_t size);
+
+/// Gets GameCardKey of size 0x134.
+Result setcalGetGameCardKey(void *key, size_t size);
+
+/// Gets GameCardCertificate of size 0x404.
+Result setcalGetGameCardCertificate(void *certificate, size_t size);
+
+/// Gets EticketDeviceKey of size 0x244.
+Result setcalGetEticketDeviceKey(void *key, size_t size);
+
 /// Converts LanguageCode to \ref SetLanguage.
 Result setMakeLanguage(u64 LanguageCode, SetLanguage *Language);
 
--- a/nx/include/switch/services/set.h
+++ b/nx/include/switch/services/set.h
@@ -319,11 +319,11 @@ typedef struct {
 } SetCalConsoleSixAxisSensorHorizontalOffset;
 
 typedef struct {
-    u32 code[0x4];
+    u32 code[0x4];          ///< Country code.
 } SetCalCountryCode;
 
 typedef struct {
-    u32 offset;                     ///< Relative to current position.
+    u32 offset;             ///< Relative to current position.
     u8 cert[0x17C];
 } SetCalEccB233DeviceCertificate;
 
@@ -336,7 +336,7 @@ typedef struct {
 } SetCalGameCardCertificate;
 
 typedef struct {
-    u32 size;           ///< Size of the entire key.
+    u32 size;               ///< Size of the entire key.
     u8 key[0x130];
 } SetCalGameCardKey;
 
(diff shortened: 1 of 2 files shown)


yellows8:
Added the structs for these to wiki, use those instead of buffer/size params.

nx/include/switch/services/set.h
```diff
@@ -271,6 +271,36 @@ void setExit(void);
 /// Gets the Service object for the actual set service session.
 Service* setGetServiceSession(void);
 
+/// Initialize setcal.
+Result setcalInitialize(void);
```
fincs: Put the new functions at the end of the file (same thing for the .c)

ITotalJustice:
Thanks for the feedback! I adding the structs, let me know if they're okay.

I noticed that all but the gamecard cert first `0x4` seems to be the offset of the actual data from after `0x4`. So data offset is `0x4 + offset`.

Also, the certs have some data in the same offsets / same values. For example the `SetCalEccB233DeviceCertificate` has the deviceID at offset `0xC4` which is also the same as the `SetCalRsa2048DeviceCertificate`.

[I have a basic example of the deviceID struct](https://github.com/ITotalJustice/NX-DeviceID-Viewer/blob/588f79f01de4840b73287c52a3e9777bdb8e5e1a/source/main.c#L15) (and also the `SetCalEccB233DeviceCertificate`). I could add this if you'd like, but just let me know how you'd like me to name it.

ITotalJustice:
I so far found two types of `NXCA`, I guess these should be an enum for the different versions?

`NintendoNXCA1Prod1` found in the eci_device_cert and ecc_b233_device_cert.
`NintendoNXCA2Prod10` found in the ssl_cert.

And lastly, looking at the gamecard cert, in the docs is says it has a cal container struct, however the first `0x4`do not appear to be a size (or an offset, like with the rest of the certs) as the number is much too large, its `0x721BE113` for me.

hexkyz:
@ITotalJustice I've updated the switchbrew wiki (https://switchbrew.org/wiki/Settings_services#set:cal) with all the sizes and names for these structs. Note that the "container" structure is only used outside of CAL0. The certificate data itself is opaque and copied directly from it's offset in calibration data, but the settings system module encapsulates this data by adding 4 extra bytes to describe it's size.

ITotalJustice:
@hexkyz Thank you for adding them, it's very helpful. I have added the structs using your information, I hope it looks okay, I will change them if not.

Yes i had noticed that sys adds `0x4` to the start of the keys, which is the size of the entire key including the '0x4' added. I have added this to the structs (that i have tested) to reflect this.

I still don't quite understand the GameCardCert. In the docs, it says the first `0x4` is the size of the cert, but the value at 0x4 does not represent this.

nx/include/switch/services/set.h
```diff
@@ -262,6 +262,128 @@ typedef struct {
...
+    u8 addr[0x6];
+} SetCalMacAddress;
+
+typedef struct {
+    u32 offset;                     ///< Relative to current position.
+    u8 cert[0x23C];
+} SetCalRsa2048DeviceCertificate;
+
+typedef struct {
+    size_t size;            ///< Size of the entire key.
```
yellows8: If this is 32bit use u32, same below (otherwise use u64).

nx/include/switch/services/set.h
```diff
@@ -885,3 +1007,33 @@ Result setsysGetTouchScreenMode(SetSysTouchScreenMode *out);
...
+/// Initialize setcal.
+Result setcalInitialize(void);
+
+/// Exit setcal.
+void setcalExit(void);
+
+/// Gets the Service object for the actual setcal service session.
+Service* setcalGetServiceSession(void);
+
+/// Gets EciDeviceCertificate of size 0x180.
```
yellows8: `Gets the \ref {full struct name}.`

nx/include/switch/services/set.h
ITotalJustice: oops! Good catch.

ITotalJustice: I have fixed this.

nx/include/switch/services/set.h
ITotalJustice: I have updated the comments to match that format.

nx/include/switch/services/set.h
```diff
@@ -262,6 +262,128 @@ typedef struct {
...
+typedef struct {
+    u8 battery_lot[0x18];
+} SetCalBatteryLot;
+
+typedef struct {
+    u8 bd_addr[0x6];
+} SetCalBdAddress;
+
+typedef struct {
+    u8 cfgId1[0x1E];
```
yellows8: `u8 cfg[...]`

nx/include/switch/services/set.h
```diff
@@ -262,6 +262,128 @@ typedef struct {
...
+typedef struct {
+    u8 offset[0x6];
+} SetCalConsoleSixAxisSensorHorizontalOffset;
+
+typedef struct {
+    u32 code[0x4];
+} SetCalCountryCode;
+
+typedef struct {
+    u32 offset;                     ///< Relative to current position.
```
yellows8: Align these comments with each other (similar to elsewhere).

ITotalJustice: I have aligned them, hope it looks okay.

hexkyz:
@ITotalJustice Oh, you mean this? https://switchbrew.org/wiki/Settings_services#GameCardCertificate
I see the confusion, the maximum size of the gamecard data is 0x400 and this is how much is read from CAL0. So it shouldn't be "it's maximum size is 0x400 bytes" but "it's maximum size is 0x404 bytes".

PR #362, opened 2020-01-02
Update documentation about fsdevMountDevice() to inform about trailing colons in device names.

DarkMatterCore:
Fixes fsdev_fixpath() returning NULL when a FsFileSystem object is mounted using a device name with a trailing colon.

### DIFF
--- a/nx/include/switch/runtime/devices/fs_dev.h
+++ b/nx/include/switch/runtime/devices/fs_dev.h
@@ -38,6 +38,7 @@ Result fsdevMountSystemSaveData(const char *name, FsSaveDataSpaceId save_data_sp
 
 /// Mounts the input fs with the specified device name. fsdev will handle closing the fs when required, including when fsdevMountDevice() fails.
 /// Returns -1 when any errors occur.
+/// Input device name string shouldn't exceed 31 characters, and shouldn't have a trailing colon.
 int fsdevMountDevice(const char *name, FsFileSystem fs);
 
 /// Unmounts the specified device.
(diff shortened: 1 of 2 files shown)


fincs:
Device names aren't supposed to have trailing colons...

DarkMatterCore:
You're right, they're not supposed to have trailing colons. However, not all things occur exactly the way we think they do. I firmly believe a safety check like this would be useful.

I have been using `fsdevGetDeviceFileSystem("sdmc:")` calls in my code to retrieve the FsFileSystem object for the SD card and they all *succeed*. This led me to believe that mounting a FsFileSystem object with `fsdevMountDevice()` and a device name with a trailing colon was safe - from my point of view, `fsdevGetDeviceFileSystem()` was taking care of the trailing colon, after all.

I was wrong, and pinpointing the exact source of the problem cost me a couple of hours. I'm just saying that, just like I made this mistake, it is very possible for someone else to get it wrong as well.

... And since there are a couple of other safety checks related to colons, another small one doesn't really hurt.

fincs:
I think a better change would have been to update the documentation.

DarkMatterCore:
If you point me in the right direction, I can gladly update the documentation myself.

fincs:
https://github.com/switchbrew/libnx/blob/master/nx/include/switch/runtime/devices/fs_dev.h#L39-L41

DarkMatterCore:
Alright then, it's done. I reverted my previous changes. Thanks for the help.

PR #373, opened 2020-02-18
Correct structs and finish adding all set:cal calls

shchmue:
Also tried to match documentation with more recent commits


nx/include/switch/services/set.h
```diff
@@ -1017,23 +1028,269 @@ void setcalExit(void);
...
+ * @brief Gets the \ref SetCalSerialNumber.
+ * @param[out] out \ref SetCalSerialNumber
+ */
+Result setcalGetSerialNumber(SetCalSerialNumber *out);
+
+/**
+ * @brief SetInitialSystemAppletProgramId
+ * @param[in] program_id input ProgramId.
+ */
+Result setcalSetInitialSystemAppletProgramId(const u64 *program_id);
```
yellows8: Don't use ptrs for input u64s when it's not arrays (same elsewhere).

shchmue: Fixed.

nx/source/services/set.c
```diff
@@ -781,6 +781,61 @@ Result setsysSetTouchScreenMode(SetSysTouchScreenMode mode) {
...
+Result setcalGetGyroscopeScale(SetCalGyroscopeScale *out) {
+    return serviceDispatchOut(&g_setsysSrv, 5, *out);
+}
+
+Result setcalGetWirelessLanMacAddress(SetCalMacAddress *out) {
+    return serviceDispatchOut(&g_setsysSrv, 6, *out);
+}
+
+Result setcalGetWirelessLanCountryCodeCount(s32 *out_count) {
+    return serviceDispatchOut(&g_setsysSrv, 7, *out_count);
```
yellows8: _setCmdNoInOutU32

nx/source/services/set.c
```diff
@@ -781,6 +781,61 @@ Result setsysSetTouchScreenMode(SetSysTouchScreenMode mode) {
...
+        .buffers = { { codes, count*sizeof(SetCalCountryCode) } },
+    );
+}
+
+Result setcalGetSerialNumber(SetCalSerialNumber *out) {
+    return serviceDispatchOut(&g_setsysSrv, 9, *out);
+}
+
+Result setcalSetInitialSystemAppletProgramId(u64 program_id) {
+    return serviceDispatchIn(&g_setsysSrv, 10, program_id);
```
yellows8: _setCmdInU64NoOut, same below.

nx/source/services/set.c
```diff
@@ -823,9 +878,156 @@ Result setcalGetGameCardCertificate(SetCalGameCardCertificate *out) {
...
+
+Result setcalGetSpeakerParameter(SetCalSpeakerParameter *out) {
+    return serviceDispatchOut(&g_setcalSrv, 22, *out);
+}
+
+Result setcalGetLcdVendorId(u32 *out_vendor_id) {
+    if (hosversionBefore(4,0,0))
+        return MAKERESULT(Module_Libnx, LibnxError_IncompatSysVer);
+
+    return serviceDispatchOut(&g_setcalSrv, 23, *out_vendor_id);
```
yellows8: _setCmdNoInOutU32

nx/include/switch/services/set.h
```diff
@@ -319,16 +327,17 @@ typedef struct {
 } SetCalConsoleSixAxisSensorHorizontalOffset;
 
 typedef struct {
-    u32 code[0x4];          ///< Country code.
-} SetCalCountryCode;
+    char code[0x3];          ///< Country code.
+} PACKED SetCalCountryCode;
```
fincs: This PACKED doesn't serve any purpose from what I can tell.

shchmue:
```c
    cal_ptr = out_cal_buffer;
    val_ptr = &out_cal_buffer->WlanCountryCodesNum;
    if ( cal_ptr->WlanCountryCodes_checksum == Crc16(&out_cal_buffer->WlanCountryCodesNum, 0x18ELL) )
    {
      code_count = *val_ptr;
      if ( code_count >= code_list_count )
        code_count = code_list_count;
      *out_count = code_count;
      if ( code_count )
      {
        _memmove_device(out_code_list, cal_ptr->WlanCountryCodes, 3LL * code_count);
        LODWORD(code_count) = 0;
      }
    }
    else
    {
      LODWORD(code_count) = 0x48C69;
    }```
From `settings` 9.0.0. It copies exactly `3 * count` bytes from `cal` to the out buffer. It is odd, but I wasn't sure how else to guarantee that, or whether this even does, but surely you'd get bad indices if we don't guarantee a 3-byte struct size. My `cal` only has one entry, as do many I imagine, so I'm not sure how to test this.

Comment (fincs) [reply nx/include/switch/services/set.h]: This structure contains an array of 3 u8s. Alignment of u8 is 1. I don't think anything is changed, or being forced to 4 byte alignment without the PACKED.

Comment (shchmue) [reply nx/include/switch/services/set.h]: you're right, i tested this in homebrew:
```c
typedef struct {
    char code[0x3];          ///< Country code.
} CountryCodeTest;

CountryCodeTest cctest[2];
printf("size of CountryCodeTest * 2 == %x\n", sizeof(cctest));
```
yields `size of CountryCodeTest * 2 == 6`

PR #376, opened 2020-02-21
add fan service

HookedBehemoth:
Command 1, 3 and 4 seam to be stubbed by now.

> [7.0.0+] SetRotationSpeedLevelByDiff/SetRotationRate no longer takes any input. GetRotationRate no longer returns output. These were presumably replaced with new cmds?


HookedBehemoth:
Adding tc might be dangerous in the hands of a fool.
What do you think about it?
>https://switchbrew.org/wiki/PTM_services#tc

yellows8:
What's the use-case for all of your currently open PRs?

HookedBehemoth:
Fan was for somebody somebody doing a monitoring overlay homebrew.
Then I was interested in tc to do manual fan tuning with the stuff from fan.

lbl was pure boredom.

yellows8:
Docs for fanSetRotationSpeedLevel and tc{Enable|Disable}FanControl should probably have a `@warning` in the docs since those could be dangerous.

nx/source/services/fan.c
```diff
@@ -0,0 +1,47 @@
...
+Result _fanInitialize(void) {
+    Result rc = smGetService(&g_fanSrv, "fan");
+    
+    if (R_SUCCEEDED(rc)) rc = _fanOpenController();
+
+    return rc;
+}
+
+void _fanCleanup(void) {
+    serviceClose(&g_fanSrv);
```
yellows8: Swap the order for these.

HookedBehemoth:
This shouldn't be merged yet.
Realized that somewhere between 6.0.0 and 9.1.0 it PTM switched to using nn::i2c::IManager::OpenSession2.
For versions before that the I2cDevice needs to be used instead of the DeviceCode.

HookedBehemoth:
found it
7.0.0+ needs 0x3d000001

hexkyz:
I think OpenController should also be exposed to the user. I believe it's more readable this way and complies with the notion of having all IPC commands delegating the handling of objects to the user.

HookedBehemoth:
Should the user handle the IController or should that still be hidden.
What about the input (PWMDevice/DeviceCode).

hexkyz:
I believe the user should handle IController. I do agree that this case is a bit odd since there's only one option for DeviceCode, but in terms of design it would fit better with the rest of services.

yellows8:
"What about the input" `u32 device_code`, for now.

nx/source/services/fan.c
```diff
@@ -0,0 +1,35 @@
...
+        .out_num_objects = 1,
+        .out_objects = &out->s,
+    );
+}
+
+Service* fanGetServiceSession(void) {
+    return &g_fanSrv;
+}
+
+Result fanSetRotationSpeedLevel(FanController *controller, float level) {
```
yellows8: fanController*, same below. Also add a fanControllerClose func.

nx/include/switch/services/fan.h
```diff
@@ -0,0 +1,32 @@
...
+
+/// Opens IController session.
+Result fanOpenController(FanController *out, u32 device_code);
+
+/// Close IController session.
+void fanControllerClose(FanController *controller);
+
+/// @warning Disabling your fan can damage your system.
+Result fanControllerSetRotationSpeedLevel(FanController *controller, float level);
+Result fanControllerGetRotationSpeedLevel(FanController *controller, float *level);
```
fincs: Missing newline at the end of file

PR #382, opened 2020-03-07
fs: be consistent about no stack temps in fsFs

SciresM:


fincs:
`fsFsSetConcatenationFileAttribute` also uses `fsFsQueryEntry`

PR #390, opened 2020-03-31
nim: add ListSystemUpdateTask/DestroySystemUpdateTask

SciresM:


nx/source/services/nim.c
```diff
@@ -0,0 +1,37 @@
+#include <string.h>
+#include "service_guard.h"
+#include "services/nim.h"
+
+#include "runtime/hosversion.h"
+
+
```
yellows8: extra newline

PR #399, opened 2020-04-14
10.0.0 Atmosphere support

SciresM:
Note: This should not be merged yet.

pgl will still need to be added to make creport function correctly.


nx/source/services/fspr.c
```diff
@@ -78,6 +78,9 @@ Result fsprSetCurrentProcess(void) {
 }
 
 Result fsprSetEnabledProgramVerification(bool enabled) {
+    if(hosversionAtLeast(10,0,0))
```
yellows8, line 81: space `if (`, same below.

PR #404, opened 2020-04-17
Add new 10.0.0 settings calls

shchmue:


yellows8:
Update wiki?

shchmue:
> Update wiki?

they're all on the wiki, only the types aren't i don't think, but as you can see i don't know anything about them as it is

yellows8:
Cmd params aren't on wiki though.

PR #487, opened 2020-06-16
Add fucntion to write to the motion controls, rather than read

shakkar23:
hopefully someone can review this, and tell me if, and why not this might work, not sure why there is only a read function, but no write one...


fincs:
HID shared memory is [read only](https://github.com/switchbrew/libnx/blob/master/nx/source/services/hid.c#L79). This PR doesn't make any sense, unfortunately. (And no, it's not possible to make it read/write, as the OS enforces that it be read only).

PR #493, opened 2020-07-12
add news service wrapper

HookedBehemoth:


nx/source/services/news.c
```diff
@@ -0,0 +1,308 @@
...
+        *total = out.total;
+    }
+
+    return rc;
+}
+
+Result newsIsSystemUpdateRequired(bool *out) {
+    u8 tmp=0;
+    Result rc = serviceDispatchOut(&g_newsSrv, 30200, tmp);
+    if (R_SUCCEEDED(rc)) *out = tmp;
```
yellows8, line 161: tmp & 1

nx/source/services/news.c
```diff
@@ -0,0 +1,308 @@
...
+
+    struct {
+        u64 current;
+        u64 total;
+    } out;
+
+    Result rc = serviceDispatchOut(&g_newsSrv, 30110, out);
+
+    if (R_SUCCEEDED(rc)) {
+        *current = out.current;
```
yellows8, line 151: `if (current) ...`, same below

nx/include/switch/services/news.h
```diff
@@ -0,0 +1,111 @@
...
+#include "../services/acc.h"
+#include "../sf/service.h"
+#include "../types.h"
+
+typedef enum {
+    NewsServiceType_Application,
+    NewsServiceType_Starter,
+    NewsServiceType_Menu,
+    NewsServiceType_DataErase,
+    NewsServiceType_v,
```
fincs, line 20: Do we know what the v actually stands for?

HookedBehemoth: Unused in official firmware npdms.

HookedBehemoth:
Server sided is good enough imo.
Do we have that for other services?

fincs:
Yes, other services such as applet have service-type checks.

HookedBehemoth:
For applet those are different interfaces.
Here it's the same exact interface with one byte difference server sided.

To answer your question: no

nx/include/switch/services/news.h
```diff
@@ -0,0 +1,111 @@
...
+ */
+
+#pragma once
+
+#include "../kernel/event.h"
+#include "../services/acc.h"
+#include "../sf/service.h"
+#include "../types.h"
+
+typedef enum {
```
SciresM, line 15:
```cpp
typedef enum {
    NewsServiceType_Administrator, // "news:a"
    NewsServiceType_Configuration, // "news:c"
    NewsServiceType_Manager,       // "news:m"
    NewsServiceType_Post,          // "news:p"
    NewsServiceType_Viewer,        // "news:v"
} NewsServiceType;
```

nx/source/services/news.c
```diff
@@ -0,0 +1,308 @@
...
+static Service g_newsSrv;
+
+static Result _newsCreateNewsService(Service *srv_out);
+
+NX_GENERATE_SERVICE_GUARD_PARAMS(news, (NewsServiceType service_type), (service_type));
+
+Result _newsInitialize(NewsServiceType service_type) {
+    Result rc;
+
+    if (service_type > NewsServiceType_v)
```
SciresM, line 18: Have NewsServiceType_Count, do >= ServiceType_Count here instead of implicitly having _Viewer as the last service type in the enum.

nx/include/switch/services/news.h
SciresM: Added a comment elsewhere with the real names, please use them.

nx/source/services/news.c
```diff
@@ -0,0 +1,308 @@
...
+
+NX_GENERATE_SERVICE_GUARD_PARAMS(news, (NewsServiceType service_type), (service_type));
+
+Result _newsInitialize(NewsServiceType service_type) {
+    Result rc;
+
+    if (service_type >= NewsServiceType_Count)
+        return MAKERESULT(Module_Libnx, LibnxError_BadInput);
+
+    const char *service_names[] = {
```
fincs, line 21:
```suggestion
    static const char* const service_names[] = {
```

Otherwise this is a mutable array that goes on the stack.

HookedBehemoth:
Are there any remaining concerns about this?

PR #496, opened 2020-09-27
psm: Add more battery functions

endrift:
Adds a few more psm functions relating to battery charging.

### DIFF
--- a/nx/include/switch/services/psm.h
+++ b/nx/include/switch/services/psm.h
@@ -40,6 +40,9 @@ Service* psmGetServiceSession(void);
 Result psmGetBatteryChargePercentage(u32 *out);
 Result psmGetChargerType(ChargerType *out);
 Result psmGetBatteryVoltageState(PsmBatteryVoltageState *out);
+Result psmGetRawBatteryChargePercentage(double *out);
+Result psmIsEnoughPowerSupplied(bool *out);
+Result psmGetBatteryAgePercentage(double *out);
 
 /**
  * @brief Wrapper func which opens a PsmSession and handles event setup.
--- a/nx/source/services/psm.c
+++ b/nx/source/services/psm.c
@@ -50,7 +50,20 @@ static Result _psmCmdInBoolNoOut(Service* srv, bool inval, u32 cmd_id) {
     return _psmCmdInU8NoOut(srv, inval!=0, cmd_id);
 }
 
+static Result _psmCmdNoInOutBool(Service* srv, bool *out, u32 cmd_id) {
+    u8 outval = 0;
+    Result rc = serviceDispatchOut(srv, cmd_id, outval);
+    if (R_SUCCEEDED(rc)) {
+        if (out) *out = outval & 1;
+    }
+    return rc;
+}
+
 static Result _psmCmdNoInOutU32(Service* srv, u32 *out, u32 cmd_id) {
+    return serviceDispatchOut(srv, cmd_id, *out);
+}
+
+static Result _psmCmdNoInOutDouble(Service* srv, double *out, u32 cmd_id) {
     return serviceDispatchOut(srv, cmd_id, *out);
 }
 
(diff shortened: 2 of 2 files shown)


nx/include/switch/services/psm.h
```diff
@@ -38,8 +38,11 @@ void psmExit(void);
 Service* psmGetServiceSession(void);
 
 Result psmGetBatteryChargePercentage(u32 *out);
+Result psmGetRawBatteryChargePercentage(double *out);
+Result psmGetBatteryAgePercentage(double *out);
 Result psmGetChargerType(ChargerType *out);
 Result psmGetBatteryVoltageState(PsmBatteryVoltageState *out);
+Result psmIsEnoughPowerSupplied(u8 *out);
```
fincs, line 45: This should be `bool *out`

endrift: Is `bool` one byte in dkA64's ABI?

fincs: Yes, it is in AAPCS64. Also note how other service wrappers tend to have xxOutBool meta-wrappers that explicitly do `value != 0`, you might want to add one here if there isn't already one.

endrift:
I have an example usage for switch-examples after this PR is merged too.

fincs (approved):

nx/source/services/psm.c
```diff
@@ -50,18 +50,43 @@ static Result _psmCmdInBoolNoOut(Service* srv, bool inval, u32 cmd_id) {
...
 
+static Result _psmCmdNoInOutDouble(Service* srv, double *out, u32 cmd_id) {
+    return serviceDispatchOut(srv, cmd_id, *out);
+}
+
 Result psmGetBatteryChargePercentage(u32 *out) {
     return _psmCmdNoInOutU32(&g_psmSrv, out, 0);
 }
 
+Result psmGetRawBatteryChargePercentage(double *out) {
```
yellows8, line 74: Func-order should be the same order as the cmdids (same in the .h).

yellows8 (approved):

CTCaer:
@endrift 
These old notes might help you:
```
x00: Input current limit u:500, a:1840, n:1200
max input current to switch for battery chg + supplement mode
gets normalized to 100, 150, 500, 900, 1200, 1500, 2000, 3000 in bq24

x04: sdp iilim? 500
x08: Charging current: 512, 2048
x0C: BatteryVoltageMax
x18: BatteryVoltageState
x1c: Battery Temp (milli C) 37589, later: 40507
x20: Capacity mAh Now: x / 20
x24: BatteryVoltageNow
x28: Capacity mAh Full: x / 20
x2C/x30: connected/charging?
x34: Charger Voltage a:9000 n:15000 (accepted contract - pdo)
x38: Charger Current a:2840 n: 2600 (accepted contract - pdo)
x3C: x1, x100, x0, x101

(u: pc, a: apple pd compat, n: nintendo)
```
I haven't looked at these since then other than pd though, so they might be wrong or need changes.
Also the following enum
https://github.com/switchbrew/libnx/blob/4afec4016d7e09ae8a3b47355a4f637d96e84d54/nx/include/switch/services/psm.h#L14-L15
is more like
1 USB-PD compatible with accepted power delivery contract
2 SDP/CDP/DCP/Type-C 1.5A/3.A non-pd/apple/qc/etc, basically any non-pd. So "usb" is right
This might help with any possible flag in the above struct.

endrift:
Yeah, I've since realized that about the enum: it's basically "does this USB-PD charger provide ample power" vs anything else.

I can confirm that a bunch of those notes are wrong, unfortunately. For example, the capacity bit is definitely wrong; that field and the other are the aforementioned raw charge and age fields but as per cent-mille instead of doubles. And some of your battery voltage etc fields are just 0 when the charger is unplugged, which doesn't make much sense.

CTCaer:
Yeah I can confirm the capacity stuff, just checked the raw dumps.
I checked one and it was dc a5 00 00 and the reported percentage as 57%, which makes sense (56.485%).

As for BatteryVoltageMax that's the bq24 configured max battery voltage (VREG:Charge Voltage Limit).
Max value when plugged in can also change depending on if battery temperature is pretty high or low.
from ptm bq limits selection:
```c
x == temperature
  temp range: state  voltage max	             current unk
     x >= 60: 4      unk == 2 ? 4208 : 3952, 2048,   0
51 >= x < 60: 3      unk == 2 ? 4208 : 3952, 2048,   unk != 2 ? 1 : 0
17 >= x < 51: 2      4208, 4208,		     1
4  >= x < 17: 1      4208, 768, 		     1
      x <  4: 0      4208, 768, 		     0
```

Battery voltage state if I remember was for the battery icon indicator level. At least that what I got, because it was changing with battery voltage (affected by no charger or charger where the battery voltage rises).
All in all it might be a flag or some other state though. (I only found dumps with 2 and 3 and bit0 was only set on plugged in dumps)

endrift:
So you're saying the age percentage is percentage of theoretical max, but can be higher than that? That seems kinda odd, though I suppose it's possible.

CTCaer:
I talked about x20 and not x28

On the other hand, capacity full / capacity design, aka age, should always be higher than 100% in new units.
(4800-4900mAh / 4667)

Even my 3 year old units sport 4750 and 4680 capacity full.

endrift:
I'm not really sure I understand. Maybe you can write up some non-shorthand docs and do a PR for that struct/service call? You seem to already understand it better than I do.

PR #501, opened 2020-11-01
Add virtmemFindAliasMemory to virtmem APIs

DCNick3:


fincs:
This won't work as intended, because `_memregionFindRandom` explicitly checks that the returned address isn't within the Alias region.

What is the intended use case of this PR? I suppose it is related to https://github.com/switchbrew/nx-hbloader/issues/35?

DCNick3:
Didn't notice that, sorry.

My intended use is as follows: I want to reserve a large region of virtual address space (probably 4 GiB) and later map physical memory to parts of it. This would allow to write efficient emulation for 32-bit systems supporting virtual memory by simply adding the reserved region base to emulated system virtual address. This is how qemu user emulation basically works.

I want it to emulate (small subset of) windows 95 APIs to run Heroes of Might and Magic III on hos.

fincs:
I see; that is indeed quite an interesting project.

We haven't tackled support for PhysicalMemory SVCs yet; but it is indeed in the cards.

fincs:
I'm closing this PR. However I'm opening an issue that is specifically about the future support for PhysicalMemory SVCs: https://github.com/switchbrew/libnx/issues/503

PR #502, opened 2020-11-02
Add service wrappers for ins:r and ins:s

averne:


fincs:
Hmm, what is this service? A bit of background would be nice to hear about before we merge this PR.

averne:
Basically it mirrors events.
Calling ins:s cmd 0 with an id 0..11 gets you a writable event handle, calling ins:r cmd 1 with an id 0..4 gets you a readable event handle.
And it will map each events according to their ids (didn't really look at that part), ie when an event from ins:s gets signalled it will signal the corresponding event(s) of ins:r.

AM calls it in idle:sys to detect whether the console recently received input, which I'd like to replicate.

yellows8:
Move into ins.h/.c?

yellows8 (approved):

nx/include/switch/services/ins.h
```diff
@@ -0,0 +1,55 @@
...
+Service* insrGetServiceSession(void);
+
+/**
+ * @brief Retrieves the last system tick the event corresponding to the ID was signaled at.
+ * @param[in] id Ins request ID (should be 0..4).
+ * @param[out] tick.
+ * @return Result code.
+ * @note The tick is only updated once per second at minimum.
+ */
+Result insrGetLastTick(u32 id, u64 *tick);
```
fincs, line 28: Is NN known to use an enum for this ID, or is it really just a plain integer?

averne: Presumably, yes.
I looked a bit more into this, IDs are mapped as follows:
- InsRequestId 0 -> InsSendId 0-11
- InsRequestId 1 -> InsSendId 0, 2-9
- InsRequestId 2 -> InsSendId 0, 2, 4-9
- InsRequestId 3 -> InsSendId 0-9, 11
- InsRequestId 4 -> InsSendId 1-9, 11

4 services retrieve ins:s write handles:
- USB: InsSendId 6
- HID: InsSendId 0-5, 9-11
- Audio: InsSendId 0
- NS: InsSendId 3, 7

Only AM subscribes to ins:r read handles, with IDs 0, 2 & 3.

From this it's not really clear to me what they are supposed to represent. I'd say low InsSendIds are used to signal activity of critical software components, while high IDs represent specific hardware (eg BT/NFC in the case of hid)?

fincs (approved):

PR #510, opened 2020-12-03
Capmtp

HookedBehemoth:
Tried to keep the initializer similar to official software.

Not sure if uint_least16_t is the correct type to use here but the uchar.h header is missing.

### DIFF
--- a/nx/include/switch/services/capmtp.h
+++ b/nx/include/switch/services/capmtp.h
@@ -9,13 +9,13 @@
 #include "../kernel/event.h"
 #include "../sf/service.h"
 
-Result capmtpInitialize(void* mem, size_t size, u32 app_count, u32 max_img, u32 max_vid, const char *other_name);
+Result capmtpInitialize(void* mem, size_t size, u32 app_count, u32 max_img, u32 max_vid, const uint_least16_t *other_name);
 void capmtpExit(void);
 
 Result capmtpStartCommandHandler(void);
 Result capmtpStopCommandHandler(void);
-Event *capmtpGetEvent1(void);
-Event *capmtpGetEvent2(void);
-Result capmtpIsRunning(bool *out);
-Result capmtpUnkBool(bool *out);
-Result capmtpGetResult(void);
+bool capmtpIsRunning(void);
+Event *capmtpGetConnectionEvent(void);
+bool capmtpIsConnected(void);
+Event *capmtpGetScanErrorEvent(void);
+Result capmtpGetScanError(void);
--- a/nx/include/switch/services/capmtp.h
+++ b/nx/include/switch/services/capmtp.h
@@ -1,0 +1,21 @@
+/**
+ * @file capmtp.h
+ * @brief capmtp service IPC wrapper.
+ * @note Only available on [11.0.0+].
+ * @author Behemoth
+ */
+#pragma once
+#include "../types.h"
+#include "../kernel/event.h"
+#include "../sf/service.h"
+
+Result capmtpInitialize(void* mem, size_t size, u32 app_count, u32 max_img, u32 max_vid, const char *other_name);
+void capmtpExit(void);
+
+Result capmtpStartCommandHandler(void);
+Result capmtpStopCommandHandler(void);
+Event *capmtpGetEvent1(void);
+Event *capmtpGetEvent2(void);
+Result capmtpIsRunning(bool *out);
+Result capmtpUnkBool(bool *out);
+Result capmtpGetResult(void);
--- a/nx/include/switch/services/capmtp.h
+++ b/nx/include/switch/services/capmtp.h
@@ -9,8 +9,11 @@
 #include "../kernel/event.h"
 #include "../sf/service.h"
 
-Result capmtpInitialize(void* mem, size_t size, u32 app_count, u32 max_img, u32 max_vid, const uint_least16_t *other_name);
+Result capmtpInitialize(void* mem, size_t size, u32 app_count, u32 max_img, u32 max_vid, const char *other_name);
 void capmtpExit(void);
+
+Service* capmtpGetRootServiceSession(void);
+Service* capmtpGetServiceSession(void);
 
 Result capmtpStartCommandHandler(void);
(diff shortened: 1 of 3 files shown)


nx/source/services/capmtp.c
```diff
@@ -0,0 +1,129 @@
...
+    if (R_SUCCEEDED(rc)) rc = smGetService(&g_capmtpRoot, "capmtp");
+    if (R_SUCCEEDED(rc)) rc = serviceConvertToDomain(&g_capmtpRoot);
+
+    if (R_SUCCEEDED(rc)) rc = _capmtpOpenSession(&g_capmtp);
+
+    if (R_SUCCEEDED(rc)) rc = _capmtpOpen(max_folders, max_img, max_vid, other_name);
+    if (R_SUCCEEDED(rc)) rc = _capmtpNoInEventOut(5, &g_connectEvent, false);
+    if (R_SUCCEEDED(rc)) rc = _capmtpNoInEventOut(7, &g_scanErrorEvent, false);
+
+    if (R_FAILED(rc)) capmtpExit();
```
yellows8, line 36: NX_GENERATE* handles this automatically.

nx/source/services/capmtp.c
```diff
@@ -0,0 +1,129 @@
...
+    const struct {
+        u32 tmem_size;
+        u32 folder_count;
+        u32 max_images;
+        u32 max_videos;
+    } in = { (u32)g_tmem.size, max_folders, max_img, max_vid };
+
+    return serviceDispatchIn(&g_capmtp, 0, in,
+        .buffer_attrs = { SfBufferAttr_In | SfBufferAttr_HipcMapAlias },
+        .buffers = { { other_name, 2*len + 1 } },
```
yellows8, line 93: Did you mean `2*(len+1)`, for 2-byte NUL-terminator, and not 1-byte?

nx/include/switch/services/capmtp.h
```diff
@@ -0,0 +1,21 @@
...
+ * @note Only available on [11.0.0+].
+ * @author Behemoth
+ */
+#pragma once
+#include "../types.h"
+#include "../kernel/event.h"
+#include "../sf/service.h"
+
+Result capmtpInitialize(void* mem, size_t size, u32 app_count, u32 max_img, u32 max_vid, const uint_least16_t *other_name);
+void capmtpExit(void);
```
yellows8, line 13: Add the usual GetServiceSession* following this?

nx/source/services/capmtp.c
```diff
@@ -0,0 +1,129 @@
...
+static Event g_connectEvent, g_scanErrorEvent;
+
+static Result _capmtpOpenSession(Service *srv);
+static Result _capmtpOpen(u32 max_folders, u32 max_img, u32 max_vid, const uint_least16_t *other_name);
+static Result _capmtpClose(void);
+static Result _capmtpNoInEventOut(u32 id, Event* event, bool autoclear);
+
+NX_GENERATE_SERVICE_GUARD_PARAMS(capmtp, (void* mem, size_t size, u32 max_folders, u32 max_img, u32 max_vid, const uint_least16_t *other_name), (mem, size, max_folders, max_img, max_vid, other_name));
+
+Result _capmtpInitialize(void* mem, size_t size, u32 max_folders, u32 max_img, u32 max_vid, const uint_least16_t *other_name) {
```
fincs, line 19: String parameters should never be anything other than utf-8. Please change this to `const char*` and handle the conversion internally.

nx/source/services/capmtp.c
```diff
@@ -0,0 +1,129 @@
...
+}
+
+static Result _capmtpOpenSession(Service *srv) {
+    return serviceDispatch(&g_capmtpRoot, 0,
+        .out_num_objects = 1,
+        .out_objects = srv,
+    );
+}
+
+static Result _capmtpOpen(u32 max_folders, u32 max_img, u32 max_vid, const uint_least16_t *other_name) {
```
fincs, line 80: Same thing here.

nx/source/services/capmtp.c
```diff
@@ -0,0 +1,129 @@
...
+    size_t len=0;
+    const uint_least16_t *tmp =other_name;
+    while(*tmp++) ++len;
```
fincs, line 83: Since this is an internal function, it shouldn't have a need to calculate the length of the string - it should just be a parameter. In addition, you'll already have the string length already calculated during the UTF-8 -> UTF-16 conversion process in the caller, so that will save time.

HookedBehemoth: What storage should be used?
Max size OFW/mtp supports is 0xff*2.

fincs: 255 Unicode codepoints (all within the BMP) fit in 255*2 bytes when encoded as utf-16, and in 255*3 bytes when encoded as utf-8.

HookedBehemoth: I meant in terms of stack, heap or user buffer?

fincs: Ah, my bad. This can easily go on stack, it's barely half a kb.

nx/source/services/capmtp.c
```diff
@@ -0,0 +1,135 @@
...
+static Result _capmtpOpenSession(Service *srv) {
+    return serviceDispatch(&g_capmtpRoot, 0,
+        .out_num_objects = 1,
+        .out_objects = srv,
+    );
+}
+
+static Result _capmtpOpen(u32 max_folders, u32 max_img, u32 max_vid, const char *other_name) {
+    u16 buffer[0x100];
+    size_t len = utf8_to_utf16(buffer, (const u8*)other_name, sizeof(buffer)/sizeof(u16));
```
fincs, line 89: Fwiw this doesn't write the null terminator. You may want to do `buffer[len] = 0` and pass `len+1` to IPC.

fincs (approved):

yellows8 (approved):

fennectech:
will this let us have MTP access to SD?

HookedBehemoth:
No

PR #518, opened 2020-12-15
fs: Added fsCreateSaveDataFileSystem(), fsDeleteSaveDataFileSystem() and fsDeleteSaveDataFileSystemBySaveDataAttribute().

DarkMatterCore:
These are needed to easily generate custom `FsSaveDataType_Cache` savedata entries.

Unlike normal savedata, cache savedata size can't be retrieved from the NACP, which means it's necessary to calculate its required size. One could then create a cache savedata entry larger than the calculated size, which is simpler than having to resize the existent save each time new data is written to it.

### DIFF
--- a/nx/include/switch/services/fs.h
+++ b/nx/include/switch/services/fs.h
@@ -88,12 +88,12 @@ typedef struct {
     u8 unused[0x190];               ///< Uninitialized.
 } FsSaveDataExtraData;
 
-/// FsSaveMeta
+/// SaveDataMetaInfo
 typedef struct {
-    u32 meta_file_size;
-    u8 meta_index;
-    u8 unk[0x0B];
-} FsSaveMeta;
+    u32 size;
+    u8 type;                        ///< \ref FsSaveDataMetaType
+    u8 reserved[0x0B];
+} FsSaveDataMetaInfo;
 
 /// SaveDataCreationInfo
 typedef struct {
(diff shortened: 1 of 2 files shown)


SciresM:
PR has mixed tabs/spaces which should be fixed before merge.

What's the use case for cache savedata for homebrew? Not necessarily objecting to merge right now (since we have other savedata creation/management stuff), but also not sure that I actually understand what this is for.

DarkMatterCore:
@SciresM Done, thanks for spotting that.

There are games like the old Doom re-releases that use cache savedata entries to store WAD packages. It's possible to replace these WAD packages to play custom levels.

nx/source/services/fs.c
```diff
@@ -224,6 +224,24 @@ Result fsOpenSdCardFileSystem(FsFileSystem* out) {
     return _fsCmdGetSession(&g_fsSrv, &out->s, 18);
 }
 
+Result fsDeleteSaveDataFileSystem(u64 application_id) {
+    const struct {
+        u64 application_id;
+    } in = { application_id };
+
+    return _fsObjectDispatchIn(&g_fsSrv, 21, in);
```
yellows8, line 232: Pass application_id directly, instead of in.

nx/include/switch/services/fs.h
```diff
@@ -88,6 +88,13 @@ typedef struct {
     u8 unused[0x190];               ///< Uninitialized.
 } FsSaveDataExtraData;
 
+/// FsSaveMeta
+typedef struct {
+    u32 meta_file_size;
+    u8 meta_index;
+    u8 unk[0x0B];
+} FsSaveMeta;
```
SciresM, line 96: This is "nn::fs::SaveDataMetaInfo", rename to FsSaveDataMetaInfo.

Rename "unk" to "reserved", since that's what the field is.

SciresM (changes requested):
FsSaveMeta should use the real typename.

nx/include/switch/services/fs.h
SciresM: Also, consider renaming "meta_index" to "meta_type", since this is nn::fs::SaveDataMetaType (enum).

Enum values are:

```
None             = 0,
Thumbnail        = 1,
ExtensionContext = 2,
```

SciresM: Honestly, rename the fields as follows:

```
meta_file_size -> size
meta_index -> type
unk -> reserved
```

DarkMatterCore:
Went ahead and added `fsDeleteSaveDataFileSystemBySaveDataAttribute()` as well. Figured it may be just as useful as `fsDeleteSaveDataFileSystem()`.

fincs (approved):

yellows8 (approved):

PR #523, opened 2021-01-18
framebuffer: Allow user override of automatic framebuffer memory allocation

SciresM:


fincs:
Superseded by 5057a191 + using NWindow API directly in ams, as previously discussed.

PR #536, opened 2021-03-18
psm: Fix charger definitions

endrift:
Update the charger definitions to match what's on the wiki: https://switchbrew.org/wiki/PTM_services#ChargerType

I'll update switch-examples with a matching fix if no changes need to be made to this PR.

### DIFF
--- a/nx/include/switch/services/psm.h
+++ b/nx/include/switch/services/psm.h
@@ -10,10 +10,11 @@
 #include "../kernel/event.h"
 
 typedef enum {
-    ChargerType_None = 0,    ///< No charger
-    ChargerType_Charger = 1, ///< Official charger or dock
-    ChargerType_Usb = 2      ///< Other USB-C chargers
-} ChargerType;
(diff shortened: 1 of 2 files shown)


nx/include/switch/services/psm.h
```diff
@@ -10,10 +10,11 @@
 #include "../kernel/event.h"
 
 typedef enum {
-    ChargerType_None = 0,    ///< No charger
-    ChargerType_Charger = 1, ///< Official charger or dock
-    ChargerType_Usb = 2      ///< Other USB-C chargers
-} ChargerType;
+    PsmChargerType_Unconnected = 0,  ///< No charger
+    PsmChargerType_EnoughPower = 1,  ///< Full speed
```
yellows8, line 14: `speed`? Was `power` intended?

endrift: I meant charging at the fastest speed supported by any wattage that could be provided, but I've fixed the phrasing to be clearer.

yellows8 (approved):

PR #545, opened 2021-04-19
Fix unused parameter warnings in tipc.h when -Wextra is used.

DarkMatterCore:


fincs:
Alternatively remove the unused parameters directly, that should be cleaner.

DarkMatterCore:
> Alternatively remove the unused parameters directly, that should be cleaner.

@fincs That would require `tipcDispatchImpl` to be modified as well. Would you prefer to remove the `TipcService` argument from it, too? TIPC macros would have to be updated as well.

fincs:
tipcDispatchImpl still needs the TipcService argument because it uses svcSendSyncRequest.

DarkMatterCore:
@fincs Just removed the unused parameters altogether.

PR #558, opened 2021-06-24
Fix IPv4 edge cases for inet_pton()

### LINKED ISSUE #557: inet_pton() failure for 0.0.0.0
It seems this library uses the same underlying logic for parsing addresses in both `inet_addr()` and `inet_pton()`. However, because `inet_addr()` interprets numbers that start with 0 as octal, this creates an incompatibility with `inet_pton()` for any address where the last byte is zero. It causes `outBase` to be returned with the value of 8, and so `inet_pton4()` reports a failure.

Since this solution for `inet_addr()` is based on libctru, perhaps it would be good to reference their version of `inet_pton()` as well.
https://github.com/devkitPro/libctru/blob/1f4669fa4c1517cc1018dc53c52662046296dff0/libctru/source/services/soc/soc_inet_pton.c#L7-L14

StephenCWills:
I've simply pulled in the libctru implementation for `inet_pton4()` to fix edge cases for IP addresses that include octets with a value of 0. I believe this should also fix some less important edge cases for IP addresses in which a mix of hexadecimal, octal, and decimal could have produced a successful result from `inet_pton()`.

This resolves #557 (tested against the DevilutionX codebase).

### DIFF
--- a/nx/source/runtime/util/inet_addr.c
+++ b/nx/source/runtime/util/inet_addr.c
@@ -1,4 +1,5 @@
 #include <stdint.h>
+#include <stdio.h>
 #include <string.h>
 #include <errno.h>
 #include <ctype.h>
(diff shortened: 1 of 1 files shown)


fincs:
See #560.

PR #560, opened 2021-06-24
Fix inet_pton by passing in base instead of getting it out

### LINKED ISSUE #557: inet_pton() failure for 0.0.0.0

endrift:
Wrote this patch yesterday but hadn't tested it yet. There was some hesitancy towards using sscanf so this may be the preferred solution over #558.

This fixes #557.

### DIFF
--- a/nx/source/runtime/util/inet_addr.c
+++ b/nx/source/runtime/util/inet_addr.c
@@ -10,7 +10,7 @@ const struct in6_addr in6addr_any = {0};
 const struct in6_addr in6addr_loopback = {.__u6_addr32 = {0, 0, 0, __builtin_bswap32(1)}};
 
 // Adapted from libctru
-static int _inetAtonDetail(int *outBase, size_t *outNumBytes, const char *cp, struct in_addr *inp) {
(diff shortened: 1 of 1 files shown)


StephenCWills:
It sounds like this has already been tested, but to be thorough, I tested this against DevilutionX just now. It worked just fine. I am a bit curious about the basis for the hesitancy towards using sscanf if someone would care to elaborate, but it's not a big deal.

fincs:
sscanf isn't exactly a cheap function at runtime, and it also has code size implications.

PR #562, opened 2021-07-20
Use DWARFv4 for IDA debugging compatibility

SciresM:
DWARFv5 is not supported by IDA, which makes getting debugging information for ELFs linking to libnx loaded impossible without this merged.


fincs:
As discussed in private, we're closing this PR.

WinterMute:
For information: The latest version of IDA now supports DWARFv5 so this is no longer necessary.

PR #564, opened 2021-07-21
nxlink: only connect to host if a valid remote address has been set.

DarkMatterCore:
This fixes a 60+ seconds block in `connect()` whenever `nxlinkConnectToHost()` is called by the NRO sent via nxlink, but the `-s` argument isn't used.

### DIFF
--- a/nx/source/runtime/nxlink.c
+++ b/nx/source/runtime/nxlink.c
@@ -7,8 +7,7 @@
 extern int __system_argc;
 extern char** __system_argv;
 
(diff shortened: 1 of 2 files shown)


DarkMatterCore:
Related to https://github.com/switchbrew/nx-hbmenu/pull/139 and https://github.com/switchbrew/switch-tools/pull/36.

WinterMute:
superceded by https://github.com/switchbrew/libnx/pull/604

PR #594, opened 2022-08-30
nifm: Update NifmWirelessSettingData to add Passphrase length

masagrator:
I don't know where NifmSfNetworkProfileData is even used, so I can't check if this miss also exists in NifmSfWirelessSettingData.


yellows8:
That's used in nifm.c.

It was already correct, passphrase[0] is the length: https://switchbrew.org/wiki/Network_Interface_services (guess this should be sync'd with wiki eventually)

PR #595, opened 2022-10-06
Fix calculation of the inline swkbd touch area

averne:
Recent versions of the sdk appear to have fixed the behavior that was reproduced previously. Below code from 11.4.3.0:
image

I took advantage of this to fix some typos


fincs:
Heh.

PR #601, opened 2023-02-01
hiddbg: modify hiddbgAttachHdlsWorkBuffer to accept a user-supplied buffer and size

ndeadly:
The current version of `hiddbgAttachHdlsWorkBuffer` is incompatible with sysmodules using libstratosphere due to the libnx memory allocations in `tmemCreate`.  This modifies the existing function to create a transfer memory from an optional user-supplied buffer using `tmemCreateFromMemory` instead.

I wasn't sure whether breaking the existing API by adding the additional parameters was a good idea, but this seems like the most straightforward solution.

### DIFF
--- a/nx/include/switch/services/hiddbg.h
+++ b/nx/include/switch/services/hiddbg.h
@@ -402,7 +402,7 @@ Result hiddbgUnsetAllAutoPilotVirtualPadState(void);
  * @brief Initialize Hdls.
  * @note Only available with [7.0.0+].
  * @param[out] session_id [13.0.0+] \ref HiddbgHdlsSessionId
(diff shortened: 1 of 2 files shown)


nx/include/switch/services/hiddbg.h
```diff
@@ -402,8 +402,10 @@ Result hiddbgUnsetAllAutoPilotVirtualPadState(void);
  * @brief Initialize Hdls.
  * @note Only available with [7.0.0+].
  * @param[out] session_id [13.0.0+] \ref HiddbgHdlsSessionId
+ * @param[in] buffer An existing buffer to be used as transfer memory. If null a new transfer memory will be created.
```
yellows8, line 405: `If null ...` <- no longer applies

yellows8 (approved):

PR #605, opened 2023-03-08
Added some definitions to /sys/socket.h

XITRIX:
Latest version of Moonlight-common-c library uses this definitions, so it is impossible to compile app without them.

To be honest I am not fully understand what they are doing, I've just copy-pasted them from Apple's /sys/socket.h file with some changes, cause it uses some Darwin types which of course not presented on Switch.

Anyway after these changes I was able to compile Moonlight it it worked perfectly.

### DIFF
--- a/nx/external/bsd/include/sys/socket.h
+++ b/nx/external/bsd/include/sys/socket.h
@@ -437,11 +437,20 @@ struct cmsghdr {
 /* followed by	u_char  cmsg_data[]; */
 };
 
-// socket credential stuff, we don't need this
-// cmsg macros, uses some obscure macro
+#define	_ALIGNBYTES		(sizeof(long) - 1)
+#define	_ALIGN(p)		(((unsigned long)(p) + _ALIGNBYTES) & ~_ALIGNBYTES)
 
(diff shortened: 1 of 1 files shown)


duckbill007:
Such macro definitions you should ad to your project and not to common library, because they not used anywhere in the library.

WinterMute:
It looks like Apple's headers are "all rights reserved" so we can't take the definitions from there. We can, however, 

> ~Such macro definitions you should ad to your project and not to common library, because they not used anywhere in the library.~

On the contrary, macro definitions contained in equivalent headers for other platforms can and should be considered for inclusion given the utility in porting work. Please don't comment here with misinformation like this.

WinterMute:
@XITRIX The Apple headers appear to be all rights reserved so we can't take definitions from there. We can take definitions from openbsd though see https://github.com/openbsd/src/blob/master/sys/sys/socket.h#L514

XITRIX:
Apple's header was almost identical to bsd's one, I've changed it a bit to be more bsd like.

But I'm not really confident about `__ALIGNBYTES32` stuff, looks like it differs between different CPU architectures, and looks like it should not be in `socket.h` header. Anyway I've compiled it to my Switch, looks like it works, but still ...

PR #606, opened 2023-04-02
tmem: Add tmemCloseHandle(), tmemWaitForPermission(), use those functions in nv.c to fix a race condition

masagrator:
When TransferMemory has different permission than RW, we need to check if `src_addr` permission was restored to original state as we currently don't have any assurance this happened when `__libnx_free` is called.

This is fixing an issue with Tesla overlays crashing randomly (mainly in docked mode) related to this bug.

Thanks to `~SciresM` for providing an idea to solve this and `~HE` from ReSwitched for help with shaping out code.


fincs:
Since this is a bug caused by a race condition in nvservices, it seems a bad idea to special case the workaround in tmemClose. An alternative solution would be to introduce a new `tmemWaitForPermission` function that would wait for a given tmem object to match the specified permission mask/value (as parameters); and use that in nv.c just before calling tmemClose.

masagrator:
This cannot be done before tmemUnmap is used, otherwise it will stuck in infinite loop.

My proposition is introducing `tmemWaitForPermissionAndClose` that would do what currently my PR is doing, but as separate function.

masagrator:
Then maybe it's a handle that must be closed first? If not, then I have no idea why it's not working.

My fork I was testing
https://github.com/switchbrew/libnx/compare/master...masagrator:libnx:patch-2

yellows8:
Impl the `tmemWaitForPermissionAndClose` you mentioned.

`Then maybe it's a handle that must be closed first? If not, then I have no idea why it's not working.` <- The handle must be closed for that to do anything.

yellows8:
Actually you can keep tmemWaitForPermission if tmemCloseHandle is added. tmemCloseHandle would have the handle-closing code from tmemClose, and you'd call this following the _nvCmdInitialize call.

masagrator:
Checked this in overlays with 2 people and it works fine.
I guess this will need grammar fixes as I'm not great at this.

yellows8:
Replace the duplicate code in tmemClose with a call to tmemCloseHandle.

nx/include/switch/kernel/tmem.h
```diff
@@ -72,6 +72,23 @@ static inline void* tmemGetAddr(TransferMemory* t){
...
 }
 
+/**
+ * @brief Closes handle of a transfer memory object.
+ * @param t Transfer memory information structure.
+ * @return Result code.
+ */
+Result tmemCloseHandle(TransferMemory* t);
+
+
```
yellows8, line 82: Remove extra newline.

nx/include/switch/kernel/tmem.h
```diff
@@ -72,6 +72,23 @@ static inline void* tmemGetAddr(TransferMemory* t){
...
+/**
+ * @brief Waits until source backing memory permissions match perm.
+ * @param t Transfer memory information structure.
+ * @param perm Permissions which the source backing memory is expected to have before return.
+ * @return Result code.
+ */
+Result tmemWaitForPermission(TransferMemory* t, Permission perm);
+
+
```
yellows8, line 91: Remove extra newline.

nx/source/services/nv.c
```diff
@@ -54,6 +54,9 @@ Result _nvInitialize(void) {
 
         if (R_SUCCEEDED(rc))
             rc = _nvCmdInitialize(CUR_PROCESS_HANDLE, g_nvTransfermem.handle, tmem_size);
+        
+        if (R_SUCCEEDED(rc))
```
yellows8, line 58: This has to be unconditional. Something like: `Result rc2 = tmemCloseHandle(...);` `if (R_SUCCEEDED(rc)) rc = rc2;`

yellows8 (changes requested):

yellows8 (approved):

fincs (approved):

masagrator:
If such hang happens, you should attach GDB to it and check if it's not stuck in loop. For example in tmemWaitForPermission().
But if this is true, then it would mean that without this function your app would crash with 0x2A8 error code.

masagrator:
I'm not sure if only RW or RW and R since this is something platform specific. 

You can try to write Perm_R instead of Perm_Rw and see what will happen, it should catch any Read permission.

PR #620, opened 2023-09-09
mutex: fix svcArbitrateLock call arguments

xerpi:
1st (thread handle) and 3rd (tag) arguments were swapped.

### DIFF
--- a/nx/source/kernel/mutex.c
+++ b/nx/source/kernel/mutex.c
@@ -55,7 +55,7 @@ void mutexLock(Mutex* m) {
         }
 
         // Ask the kernel to arbitrate the lock for us.
-        if (UNLIKELY(R_FAILED(svcArbitrateLock(value & ~HANDLE_WAIT_MASK, m, cur_handle)))) {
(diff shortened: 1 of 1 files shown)


fincs:
https://github.com/Atmosphere-NX/Atmosphere/blob/master/libraries/libmesosphere/source/svc/kern_svc_lock.cpp#L28

The first argument is the handle of the thread that is currently holding the lock, not our own thread handle. The third argument is our own tag (not an actual HOS thread handle). This tag is just the address to the current thread's thread-local region, which is retrieved a bit earlier: https://github.com/switchbrew/libnx/blob/master/nx/source/kernel/mutex.c#L34C5-L34C38 The `cur_handle` name on the variable holding the tag might have misled you.

xerpi:
> https://github.com/Atmosphere-NX/Atmosphere/blob/master/libraries/libmesosphere/source/svc/kern_svc_lock.cpp#L28
> 
> The first argument is the handle of the thread that is currently holding the lock, not our own thread handle. The third argument is our own tag. This tag is the value that is supposed to be our own thread handle, which is retrieved a bit earlier: https://github.com/switchbrew/libnx/blob/master/nx/source/kernel/mutex.c#L34C5-L34C38

That makes more sense, thanks for the explanation!

PR #629, opened 2023-10-25
erpt: add FS commands needed for SubmitFsInfo()

SciresM:
These are all needed for latest 17.x changes in atmosphere.

### DIFF
--- a/nx/include/switch/services/fs.h
+++ b/nx/include/switch/services/fs.h
@@ -350,9 +350,9 @@ typedef struct {
 
 /// FatFatError
 typedef struct {
-    int error;
-    int extra_error;
-    int drive_id;
+    s32 error;
+    s32 extra_error;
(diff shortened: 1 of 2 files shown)


nx/include/switch/services/fs.h
```diff
@@ -340,6 +340,108 @@ typedef enum {
...
+typedef struct {
+    u32 num_activation_failures;
+    u32 num_activation_error_corrections;
+    u32 num_read_write_failures;
+    u32 num_read_write_error_corrections;
+} FsStorageErrorInfo;
+
+/// FatFatError
+typedef struct {
+    int error;
```
yellows8, line 353: int -> s32

SciresM: Done

yellows8 (approved):

PR #635, opened 2023-11-23
add fsDeviceOperatorGetGameCardUpdatePartitionInfo

ITotalJustice:
i wasn't sure if the `version` field should be a bitfield or just a u32. if we keep it a u32, should i comment how to extract the version as it might not be obvious to the dev, as they may try to use `HOSVER_X` macros on it

```c
const u8 relstep = (update_partition_info.version >> 0)  & 0xFFFF;
const u8 micro   = (update_partition_info.version >> 16) & 0x000F;
const u8 minor   = (update_partition_info.version >> 20) & 0x003F;
const u8 major   = (update_partition_info.version >> 26) & 0x003F;
```

### DIFF
--- a/nx/include/switch/services/fs.h
+++ b/nx/include/switch/services/fs.h
@@ -266,6 +266,12 @@ typedef struct {
 } FsGameCardHandle;
 
 typedef struct {
+    u32 version;
+    u8 pad[0x4];
+    u64 tid;
+} FsCardUpdatePartitionInfo;
+
+typedef struct {
     u32 aes_ctr_key_type;           ///< Contains bitflags describing how data is AES encrypted.
     u32 speed_emulation_type;       ///< Contains bitflags describing how data is emulated.
     u32 reserved[0x38/sizeof(u32)];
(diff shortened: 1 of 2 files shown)


nx/include/switch/services/fs.h
```diff
@@ -265,6 +265,12 @@ typedef struct {
     u32 value;
 } FsGameCardHandle;
 
+typedef struct {
+    u32 version;
+    u8 pad[0x4];
+    u64 tid;
+} FsCardUpdatePartitionInfo;
```
yellows8, line 272: FsGameCardUpdatePartitionInfo

nx/include/switch/services/fs.h
```diff
@@ -265,6 +265,12 @@ typedef struct {
     u32 value;
 } FsGameCardHandle;
 
+typedef struct {
+    u32 version;
+    u8 pad[0x4];
+    u64 tid;
```
yellows8, line 271: id

yellows8:
`u32 version` is fine, `u32 version` is used elsewhere.

yellows8 (approved):

PR #637, opened 2023-11-30
nvchannel: Fix SET_CLK_RATE, implement GET_CLK_RATE and SET_SUBMIT_TIMEOUT

averne:
The version check was accidentally inverted in SET_CLK_RATE.
These are used for video engines and VIC.

### DIFF
--- a/nx/include/switch/nvidia/ioctl.h
+++ b/nx/include/switch/nvidia/ioctl.h
@@ -162,8 +162,8 @@ typedef struct {
 } nvioctl_command_buffer_map;
 
 typedef struct {
-    uint32_t rate;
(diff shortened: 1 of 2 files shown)


nx/include/switch/nvidia/ioctl.h
```diff
@@ -161,6 +161,11 @@ typedef struct {
     u32 iova;
 } nvioctl_command_buffer_map;
 
+typedef struct {
+    uint32_t rate;
+    uint32_t moduleid;
```
fincs, line 166:
```suggestion
    u32 rate;
    u32 moduleid;
```

PR #640, opened 2024-02-20
ipv6: in.h extended

XITRIX:
To be able to use [MDNS lib](https://github.com/mjansson/mdns) I had to add header for ipv6 address structure

### DIFF
--- a/nx/external/bsd/include/netinet/in.h
+++ b/nx/external/bsd/include/netinet/in.h
@@ -684,6 +684,14 @@ struct sockaddr_in6 {
 	uint32_t        sin6_scope_id; /* Scope ID */
 };
 
+/*
+ * Argument structure for IPV6_JOIN_GROUP and IPV6_LEAVE_GROUP.
+ */
+struct ipv6_mreq {
+	struct in6_addr ipv6mr_multiaddr;
+	unsigned int    ipv6mr_interface;
+};
+
 extern const struct in6_addr in6addr_any;
 extern const struct in6_addr in6addr_loopback;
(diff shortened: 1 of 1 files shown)


WinterMute:
fwiw the switch network stack doesn't support ipv6 so there isn't really much point in dns providing ipv6 addresses. You might be better off looking at making ipv6 support optional in that lib instead.

XITRIX:
Yes, it does not support it, but in terms of portability I'd prefer to be able to compile non modified version of library, than forking it and doing any modifications to rip out functionality for ipv6 just for Switch. I'm making cross platform app, so it will be a really painful if I'll need to do it for each platform like this, than just add missing headers and allow it to build without any problems

WinterMute:
> Yes, it does not support it, but in terms of portability and maintainability I'd prefer to be able to compile non modified version of library, than forking it and doing any modifications just for Switch. I'm making cross platform app, so it will be a really painful if I'll need to do it for each platform like this, than just add missing headers and allow it to build without any problems

The switch isn't the only system around which doesn't support ipv6. It would seem reasonably to upstream a patch there to allow building without ipv6 support. That would be my recommended option here rather than adding things to libnx headers.

XITRIX:
Maybe you are right, I just saw some structures for ipv6 already presented in switch headers, and this was the only one missing to be able to successfully compile and run this lib for Switch, but I just checked the solution for this lib on PSVita, and as you said, they forked this library and added IPV6_UNSUPPORTED definition for vita around ipv6 related functions

PR #641, opened 2024-03-29
Basic support for 18.0.0

SciresM:
This adds version checking to removed commands, and adds the new info types (we apparently never added 16.0.0's IoRegionHint).

The hidsys commands can be cloned to hiddbg, not sure how that should be handled though?

### DIFF
--- a/nx/source/services/hidsys.c
+++ b/nx/source/services/hidsys.c
@@ -497,14 +497,14 @@ Result hidsysLegacyIsButtonConfigSupported(HidsysUniquePadId unique_pad_id, bool
 }
 
 Result hidsysIsButtonConfigSupported(BtdrvAddress addr, bool *out) {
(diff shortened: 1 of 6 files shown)


nx/source/services/hidsys.c
```diff
@@ -497,14 +497,14 @@ Result hidsysLegacyIsButtonConfigSupported(HidsysUniquePadId unique_pad_id, bool
 }
 
 Result hidsysIsButtonConfigSupported(BtdrvAddress addr, bool *out) {
-    if (hosversionBefore(11,0,0))
+    if (hosversionBefore(11,0,0) || hosversionAtLeast(18,0,0))
```
yellows8, line 500: hosversionBetween

yellows8 (approved):

PR #654, opened 2024-11-26
Update setsysEdid struct

masagrator:
At this moment `setsysGetEdid()` and `setsysSetEdid()` are failing because struct size (0x100) is smaller than expected size (0x200). 
This PR fixes use of those functions. 

Also added note for manufacture year because result was always 10 years ahead of what was supposed to be.

### DIFF
--- a/nx/include/switch/services/set.h
+++ b/nx/include/switch/services/set.h
@@ -497,7 +497,7 @@ typedef struct {
     u16 product_code;
     u32 serial_number;
     u8 manufacture_week;
-    u8 manufacture_year;
+    u8 manufacture_year;                    ///< Real value is val - 10.
     u8 edid_version;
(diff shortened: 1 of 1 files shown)


nx/include/switch/services/set.h
```diff
@@ -560,6 +560,7 @@ typedef struct {
     SetSysModeLine extended_timing_descriptor[5];
     u8 padding[5];
     u8 extended_checksum;                   ///< Sum of 128 extended bytes should equal 0 mod 256.
+    char reserved[0x100];
```
yellows8, line 563: Replace with: `u8 data2[0x80]; ///< [13.0.0+]` `u8 data3[0x80]; ///< [13.0.0+]`

yellows8 (approved):

PR #668, opened 2025-04-26
Add nxlinkConnectToHost note about the socket driver

alloncm:
Add a comment that the socket driver must be initialized prior to calling this function, otherwise this function will fail.

After debugging it for sometime I found [this post] (https://gbatemp.net/threads/my-sdl2-game-crash-at-launch-how-to-debug-it.611957/) which helped me and I hope this could save some time for others as well.


masagrator:
socket must be initialized for any kind of server-client interaction in network. That's not limited to this function, and that's not limited to this platform.

alloncm:
You are correct I just assumed that it will be mentioned in the docs / have an indicative error message / the function will initialize it itself (though this might be risky on second thought).

Perhaps setting `errno` to an indicative error number would be better?

Anyway this is my first time developing for the switch so feel free to close this if you think it is trivial.

PR #669, opened 2025-04-29
btdrv: document timeout parameter to btdrvTriggerConnection

ndeadly:


masagrator:
Is it possible to explain what time measure is used for timeout?

ndeadly:
> Do we know which time unit is used for timeout?

It's not really clear. If you query the BCM4356 for the TBFC parameters used internally for this function the default host trigger timeout value is 0x1800/6144, which doesn't immediately jump out as some obvious round interval. Is a Kibisecond a valid unit of measurement? XD

yellows8 (approved):

masagrator:
Since this is BLE, maybe it's `timeout * 625us`? Then 0x1800 gives us 3.84 seconds.

ndeadly:
I considered that, but 3.84 is a rather specific number.

> Is a Kibisecond a valid unit of measurement? XD

Maybe this wasn't so dumb after all. It might literal "time units"...
 https://en.wikipedia.org/wiki/TU_(time_unit)

PR #675, opened 2025-05-17
nfp: Improve types and enums

XorTroll:
This improves naming on some structs/fields, and fixes others which had wrong definitions (NfpData currently has the wrong size due to using the wrong mii type).
Removed some DeviceState values that are not exposed through commands, and merged State (all services internally use the same state values).

### DIFF
--- a/nx/include/switch/services/nfc.h
+++ b/nx/include/switch/services/nfc.h
@@ -108,6 +108,12 @@ typedef enum {
     NfpAmiiboFlag_ApplicationAreaExists = BIT(1),   ///< Application area exists.
 } NfpAmiiboFlag;
 
+typedef enum {
+    NfpBreakType_Flush  = 0,
+    NfpBreakType_Break1 = 1,
+    NfpBreakType_Break2 = 2,
+} NfpBreakType;
+
 typedef struct {
     u16 year;
     u8 month;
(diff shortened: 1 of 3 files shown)


nx/include/switch/services/nfc.h
```diff
@@ -102,103 +103,120 @@ typedef enum {
...
-    u8  uuid[10];
-    u8  uuid_length;
-    u8  reserved1[0x15];
-    u32 protocol;
-    u32 tag_type;
-    u8  reserved2[0x30];
+    u16 year;
+    u8 month;
+    u8 day;
+} NX_PACKED NfpDate;
```
yellows8, line 121: Remove NX_PACKED where it's not needed.

XorTroll: Done, checked that sizes are now fine with static_assert tests

nx/include/switch/services/nfc.h
```diff
@@ -102,104 +103,123 @@ typedef enum {
...
 typedef struct {
-    u8 amiibo_id[0x8];
-    u8 reserved[0x38];
-} NX_PACKED NfpModelInfo;
+    union {
+        u8 character_id[3];
+        struct {
+            u16 game_character_id;
+            u8 character_variant;
+        } NX_PACKED;
```
yellows8, line 157: Is NX_PACKED needed here?

XorTroll: Yeah, otherwise the struct would have a size of 4 bytes

yellows8 (approved):

PR #679, opened 2025-06-02
remove unused version number in Makefile

WinterMute:
Supercedes https://github.com/switchbrew/libnx/pull/678

### DIFF
--- a/nx/Makefile
+++ b/nx/Makefile
@@ -98,6 +98,9 @@ install: lib/libnx.a lib/libnxd.a
 	@cp -rv include $(DESTDIR)$(DEVKITPRO)/libnx/
 	@cp -rv external/bsd/include $(DESTDIR)$(DEVKITPRO)/libnx/
 
+dist-src:
+	@tar -cjf libnx-`git describe --tags | sed 's/^v//'`.tar.bz2 include source data external Makefile default_icon.jpg switch_rules switch.ld switch.specs
+
 #dox:
 #	@doxygen Doxyfile
 #	@doxygen Doxyfile.internal


averne:
Hello, personally I appreciate that the current script creates an archive of the compiled code during the install step, which can be useful for regression tracking.
Could you consider leaving this in?

WinterMute:
> Hello, personally I appreciate that the current script creates an archive of the compiled code during the install step, which can be useful for regression tracking. Could you consider leaving this in?

Have you considered using git for this purpose?

SciresM:
I would advocate for choosing #678 over this myself. I have to install wacky libnx versions myself and manage them without git commits, and would appreciate us keeping the dist-src target. I will probably pull request to re-add it if this is merged.

WinterMute:
An unused version number in the Makefile is a maintenance burden which doesn't need to exist. This is just cruft that was carried over from before we maintained everything via pacman.

WinterMute:
> (Automatically parsing the commit hash the way ams does would, I believe, completely solve the "maintenance burden" problem.)

git hashes aren't available when packaging, archives are obtained from the github tag based on the version number in the PKGBUILD.

nx/Makefile
```diff
@@ -98,14 +91,15 @@ all: lib/libnx.a lib/libnxd.a
 dist-bin: all
 	@tar --exclude=*~ -cjf libnx-$(VERSION).tar.bz2 include lib default_icon.jpg switch_rules switch.ld switch.specs -C external/bsd include
```
oreo639, line 92: dist-bin still references the removed `VERSION`.

PR #715, opened 2026-02-20
Update comment for nsGetApplicationControlData2

masagrator:


DarkMatterCore:
Given what we now know, I think it'd be fair and more accurate to rename `flag2` to `acd_idx` and `flag1` to just `flag`. The requested value for the `flag2` argument matches the `u8` at `0x3216` within the returned NACP (`1` being used by the NACP at `/Acd_0/control.nacp`).

nx/include/switch/services/ns.h
```diff
@@ -320,7 +320,7 @@ Result nsGetApplicationControlData(NsApplicationControlSource source, u64 applic
  * @param[out] buffer \ref NsApplicationControlData
  * @param[in] size Size of the buffer.
  * @param[in] flag1 Default is 0. 0xFF speeds up execution.
- * @param[in] flag2 Default is 0.
+ * @param[in] flag2 Default is 0. 1 tries to get NsApplicationControlData for Switch 2 Edition.
```
yellows8: Rename flag2 to acd_index.

yellows8 (approved):

nx/include/switch/services/ns.h
masagrator: Done

PR #721, opened 2026-04-11
Update hid header to include information about applet services requirement

masagrator:
Was trying to figure out why I can't start sixaxis sensor in my overlay, and it turns out because in hid.c it uses applet call that needs initialized applet to return correct value.

This is my attempt at trying to better expose this requirement where it's necessary.


nx/include/switch/services/hid.h
```diff
@@ -1485,6 +1485,7 @@ void* hidGetSharedmemAddr(void);
 ///@{
 
 /// Initialize TouchScreen. Must be called when TouchScreen is being used. Used automatically by \ref hidScanInput when required.
+/// Applet services must be initialized.
```
yellows8: Most hid cmds take PID+aruid - so moving these to just a `@note` after the `@file ... @brief` would make more sense.

masagrator: Done.

yellows8 (approved):

PR #722, opened 2026-04-13
Update nsGetApplicationControlData2 informations 

masagrator:


nx/include/switch/services/ns.h
```diff
@@ -319,7 +319,7 @@ Result nsGetApplicationControlData(NsApplicationControlSource source, u64 applic
  * @param[in] application_id ApplicationId.
  * @param[out] buffer \ref NsApplicationControlData
  * @param[in] size Size of the buffer.
- * @param[in] flag1 Default is 0. 0xFF speeds up execution.
+ * @param[in] flag1 Default is 0. 1 retrieves 174x174 icon instead of 256x256. 0xFF uses alternative path that is faster on 19.0.0-21.2.0, but slower on 22.0.0+ than 0 and 1.
```
yellows8: Would prefer what 0xFF actually does being listed here, `an unknown alternative path.` otherwise.

yellows8 (approved):

nx/include/switch/services/ns.h
masagrator: I know only that it skips some additional functions that are used to parse data from some specific type of cartridges, that's why it fails to read very few cartridge games without update installed in sdcard/nand. Added unknown.

PR #728, opened 2026-04-25
Add "idle:sys" service

masagrator:


yellows8:
Is this for use from a context where applet is unusable (sysmodule)?

masagrator:
Yes, there are already some overlays which reimplemented this to block Switch from sleeping when overlay is active.

yellows8 (approved):

nx/include/switch/services/idlesys.h
```diff
@@ -0,0 +1,21 @@
...
+Result idlesysInitialize(void);
+
+/// Exit idle:sys.
+void idlesysExit(void);
+
+/// Gets the Service for idle:sys.
+Service* idlesysGetServiceSession(void);
+
+/// Resets sleep counter.
+Result idlesysReportUserIsActive();
```
fincs: `(void)`

nx/source/services/idlesys.c
```diff
@@ -0,0 +1,23 @@
...
+
+void _idlesysCleanup(void) {
+    serviceClose(&g_idlesysSrv);
+}
+
+Service* idlesysGetServiceSession(void) {
+    return &g_idlesysSrv;
+}
+
+Result idlesysReportUserIsActive() {
```
fincs: `(void)`

fincs (changes requested):

fincs (approved):

PR #734, opened 2026-06-28
add clkrstSetMinVClockRate() and clkrstGetDvfsTable()

Lightos1:


nx/source/services/clkrst.c
```diff
@@ -61,3 +61,36 @@ Result clkrstGetPossibleClockRates(ClkrstSession *session, u32 *rates, s32 max_c
...
+    );
+
+    if (R_SUCCEEDED(rc) && out_count) {
+        *out_count = out;
+    }
+
+    return rc;
+}
+
+Result clkrstSetMinVClockRate(ClkrstSession *session, u32 hz) {
```
yellows8: Cmds should be in cmdid-order. Move this before clkrstGetPossibleClockRates.

yellows8: This should also be clkrstSetMinimumVoltageClockRate.

yellows8 (approved):

Lightos1:
Realized that the firmware checks didn't make much sense since clkrst was only added after 8.0.0 anyway, I accidentally confused it with the older pcv versions, my bad.

PR #735, opened 2026-07-27
set: Add setfdSetSettingsItemValue and setfdResetSettingsItemValue

ZachyCatGames:


nx/include/switch/services/set.h
```diff
@@ -2624,3 +2624,25 @@ Result setcalGetDeviceId(u64 *out_device_id);
  * @param[out] out_type Output ConsoleSixAxisSensorMountType.
  */
 Result setcalGetConsoleSixAxisSensorMountType(u8 *out_type);
+
+// Initialize setfd.
```
yellows8: Add comment: `[4.0.0+] Only exposed if in development mode.`

ZachyCatGames: done

yellows8: For doxygen these also need to be `///`.

ZachyCatGames: got it, also added `setfdGetServiceSession` since I noticed I missed it before.

yellows8 (approved):

souldbminerr:
Is the "development mode" mentioned here the one that Atmosphere activates? or the actual one on i.e devkits?

